PR #1, opened 2016-05-24
Parallel make & bugfixes

scaronni:
Hello,

I'm the Fedora/RHEL maintainer for the DKMS package. As you can see from the commit history I've pushed some stuff to the original DKMS repository while integrating things in Fedora/RHEL, especially all the dependency logic required for ZFS modules.

Here are three other rebased patches for Fedora issues that have already been solved in the old Debian  DKMS package.

Would it be possible to have commit access to the repository directly?

Thanks,
--Simone


jared-dominguez:
I'll merge this. Thanks for your contributions!

And yes, co-maintainers are welcome. :)

PR #5, opened 2016-07-21
Fix version check if in-kernel modinfo is missing 'version:' tag

bollie:
- Some kernel modules (openvswitch.ko) do not have a 'version:' tag.
  In this case get_module_verinfo() returns res[0] empty.
  check_version_sanity() flattens the array promoting the
  'srcversion:' tag to the primary version.
- This means that if a module with a version tries to replace a
  module without a version, --force is needed. This is not ideal.
- This patch assures that a version is always present in the array.
  If the tag is missing, the version is set to '0'.

### DIFF
--- a/dkms
+++ b/dkms
@@ -709,6 +709,7 @@ read_conf()
 get_module_verinfo(){
     unset res
     local vals=
+    res[0]=0 #if "version:" tag is missing, assume it is 0
     while read -a vals; do
     case ${vals[0]} in
         version:)


jared-dominguez:
First, thank you for contributing!

Here are my thoughts on this PR. A version really ought to be provided by the module. I think that is the most reasonable thing to do, especially since there's no such thing as a stable kernel ABI in the Linux world. If the upstream maintainer for the kernel module is not willing to do that, I think an acceptable alternative is for the distribution maintainer to version the module.

Short of achieving the above, I'd entertain some DKMS change that at absolute minimum logs a warning, but I think more is likely required than just a log entry.

bollie:
I agree wholeheartedly that a version tag should be provided by the module, but in this case the module without the version is the in-tree shipped module!

I was assuming that the check_version_sanity script tried to prevent modules from being downgraded. In this respect, what I expected to happen was that a module with a version tag (i.e. the one just built by DKMS) would override the in-tree kernel module without the tag.

I understand that a blanket "assume non-tagged modules are at a lower version than tagged modules" would mean that DKMS will keep overriding modules until they get a version.

In the case of absolute pedanticness, the check_version_sanity script should fail if either module is unversioned. BUT, for the sake of pragmatism, I feel that an override mechanism should be available for the DKMS package builder (in dkms.conf, for example).

jared-dominguez:
@bollie I am hesitant to make the assumption that in-tree trumps correct behavior. I think openvswitch's version behavior should be treated as a bug needing to be fixed. Barring that fix being accepted, my preference still is for distributions to handle versioning, especially considering that you very likely need to version your distribution's DKMS package so could use that version.

jared-dominguez:
Closing as there has been no movement in months. Feel free to open a new pull request with an implementation following the guidelines I suggested.

sspans:
This is affecting us as well, we are currently hot-patching dkms as a workaround.
Given that this affects a large set of in the wild distro kernels it would make sense to add something to DKMS.

PR #16, opened 2017-01-10
Adding support for building suse rpm with systemd

pliniosilveira:
Adding support for building suse rpm with systemd

Reference:
https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines


jared-dominguez:
Looks good to me.

PR #21, opened 2017-04-13
postinst: Fix newest kernel detection

iainlane:
We were passing a list of kernels to get_newest_kernel(), when it
expected a single kernel as its first argument - the currently running
kernel. This in turn meant that the sometimes (if the first thing in the
list was itself the newest kernel), the wrong kernel was passed into
_get_newest_kernel_debian(). As a result this falsely reported that
there was no 'newest' kernel available and then didn't build modules for
it.

Drop the parameter to get_newest_kernel(), and instead rely on
CURRENT_KERNEL being set to $(uname -r).

https://bugs.launchpad.net/ubuntu/+source/dkms/+bug/1681566

### DIFF
--- a/dkms_common.postinst
+++ b/dkms_common.postinst
@@ -114,7 +114,6 @@ get_newest_kernel() {
     # Try Debian first as rpm can be installed in Debian based distros
     if [ -e /usr/bin/dpkg ]; then
         # If DEB based
-        CURRENT_KERNEL=$1
         CURRENT_VERSION=${CURRENT_KERNEL%%-*}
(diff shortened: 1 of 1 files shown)


iainlane:
Please see the LP bug - we were noticing that dkms modules weren't being built for the new kernels on yakkety -> zesty. I'm not totally sure this fix works for all usecases, so I would appreciate a review.

It is something we would like to SRU ASAP if possible, so that upgraders get modules built.

yuzaipiaofei:
Looks good to me .

Issue #2, opened 2016-05-24
Backing up current initrd can fill up partitions


scaronni:
According to: https://github.com/dell-oss/dkms/blob/master/dkms#L267-L283

An old initrd is saved with the "old-dkms" suffix at every rebuild, to revert to the previous initrd in case of problems. This is fine, but it can lead to a full /boot partition after a few update, as old initrd images are never deleted (ref: https://bugzilla.redhat.com/show_bug.cgi?id=1251602).

What is the best approach here? Is it ok if I create a patch that deletes all "old-dkms" images prior to starting a new initird build? If a new initird build is triggered, this basically means that the user is updating the kernel or the module, and in both cases it ensures us that the "current boot" is fine and there is no need to revert.

What do you think?

Thanks,
--Simone

jared-dominguez:
Ideally old initrds for still installed kernels should be kept around. I'm fine accepting a cleanup patch that saves those.

nightsky30:
Affecting Debian/Ubuntu as well:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=717584
https://bugs.launchpad.net/ubuntu/+source/dkms/+bug/1515513
https://bugs.launchpad.net/ubuntu/+source/unattended-upgrades/+bug/1357093

jarnos:
https://github.com/dell/dkms/pull/22

jarnos:
Backup files could be cleared in update-initramfs's delete function, too.

yuzaipiaofei:
@scaronni Thanks your code work.

murraybd:
This is actually incorrect as the remove_initrd_backup command runs before the dkms remove command which actually makes the .old-dkms file.

jarnos:
Why does it make a .old-dkms file when it runs dkms remove?

murraybd:
Because do_uninstall calls make_initrd and dkms, as far as I can tell, doesn't know the kernel is scheduled for removal just that the module is being removed so it remakes the initrd even though its getting removed.

jarnos:
Well, wouldn't it be better to tell make_initrd that kernel is being removed so there is no need to create a backup?

PR #22, opened 2017-04-13
Remove possible .old-dkms initrd backup

jarnos:
Fixing https://github.com/dell/dkms/issues/2

### DIFF
--- a/dkms
+++ b/dkms
@@ -286,7 +286,6 @@ make_initrd()
 
     # Back up our current initrd
     echo $""
-    [[ $2 = ia64 && -d /boot/efi/efi/redhat ]] && initrd_dir="/boot/efi/efi/redhat"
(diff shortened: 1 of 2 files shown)


superm1:
In principle this makes sense.  The stuff to remove the odd redhat path is an odd bit from itanium.  Really I don't know any reason that itanium should still have a special codepath in DKMS these days.

Redhat even claims that Itanium is unsupported these days: https://access.redhat.com/articles/rhel-limits

Would you add a second commit that removes the RH specific path from this prerm.d as well as any other references?  Should be in good shape to merge then.

superm1:
@yuzaipiaofei Looks good to me now.  Please review and merge.

jarnos:
Changed dkms file respectively in the latter commit.

PR #23, opened 2017-05-22
Red hat cleanups

scaronni:
I have commit access to the repository but would like to have some feedback on this.

This is a pull request for a few things:
- Remove UDEV_TRIGGER information from the man page as it has been deleted long ago
- Make sure on systemd systems that new modules are loaded after being built (rationale in the commit message)
- Remove stuff for old end of life Red Hat / Fedora versions.

I don't know how the policy for DKMS is; if you still want to support "dead" distributions for a while or not. Personally, I don't see the point for doing that. As the EPEL/Fedora maintainer of DKMS I'm not be able to push any new change to those EOL distributions.

Thanks,
--Simone

### DIFF
--- a/dkms
+++ b/dkms
@@ -1526,6 +1526,12 @@ install_module()
             exit 6
     }
 
+    # Make the newly installed modules available immediately
+    find /sys/devices -name modalias -print0 | xargs -0 cat | xargs modprobe -a -b -q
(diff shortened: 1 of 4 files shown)


jared-dominguez:
It all looks good to me. Thanks for the code cleanup!

PR #24, opened 2017-06-08
Print correct path to dkms.conf when it was not found

vden:
I've run into the issue when I had accidentally removed one of directories in dkms tree.
Then `dkms status` started to fail with non-informative message `File:    does not exist.`

This small patch fixes this issue showing actual path to file which dkms failed to read.


yuzaipiaofei:
Looks Good .

PR #27, opened 2017-07-08
Revert "Make newly installed modules available immediately"

seblu:
This reverts commit f5bfb12fef1fc06e56355cdba500eaa98d4e6aa8.

Modules installation should install modules, not load them. It creates issues when you build modules for non running kernels or when you do reload in a middle of a bunch of modules rebuild.
Moreover Arch Linux is not doing modules automatically.

See https://bugs.archlinux.org/task/54481

### DIFF
--- a/dkms
+++ b/dkms
@@ -1522,12 +1522,6 @@ install_module()
             exit 6
     }
 
-    # Make the newly installed modules available immediately
-    find /sys/devices -name modalias -print0 | xargs -0 cat | xargs modprobe -a -b -q
-    if [ -f /lib/systemd/system/systemd-modules-load.service ]; then
-        systemctl restart systemd-modules-load.service
-    fi
-
     # Do remake_initrd things (save old initrd)
     [[ $remake_initrd ]] && ! make_initrd "$kernelver" "$arch" && {
         do_uninstall "$kernelver" "$arch"


jwrdegoede:
Hi,

Author of the commit this reverts here. Looking at the archlinux bugreport it seems to me that the problem is the systemctl restart systemd-modules-load.service call. I added that for completeness sake, but the important call to make sure hardware related modules are loaded when build during system bootup really is the find /sys/devices -name modalias ... line.

IMHO a plain revert is a too big hammer. Can you please submit a patch instead just removing these 3 lines :  ?

if [ -f /lib/systemd/system/systemd-modules-load.service ]; then
    systemctl restart systemd-modules-load.service
fi

That would be fine with me and not cause regressions for Fedora users.

Regards,

Hans

seblu:
Hi @jwrdegoede,

The real issue is that you want to **load** modules when you **installing** them in the filesystem tree. This is different concept, like **build** and **install**.

It's not expected when you build or install modules for another kernel that the running one, to try to load all system modules. Neither when you are updating an already loaded module.
Note that restart `systemd-modules-load.service` does not remove and add already loaded modules, so calling it is useless. 
And when there is modules with dependency, restarting the `systemd-modules-load.service`  display failure because it's too soon.

Despite dkms has a several hardcoded path, `/lib/systemd/system/systemd-modules-load.service` may change based on systemd installation.

If you want add a logic for fedora please propose a patch which detect fedora and act only for fedora or allow a switch to enable/disable the loading behavior (systemd has a larger audience).
Even for Fedora I don't get why dkms is the place to load/reload all modules defined in `modules-load`. Make more sens to handle it once in a post install hook in the package manager.

superm1:
When modules are built during bootup under Fedora, what service is doing this task?  It's dkms_autoinstaller right?  If the main benefit is to have them available immediately if done during bootup, I think that it would make sense to move that code to dkms_autoinstaller.

seblu:
It's not wise to build modules during bootup; you may need some of them before being able to setup a proper build environment (e.g zfs modules). Best moment is to build after installation when you are on it.

superm1:
@seblu I generally agree.  @jwrdegoede why are the modules built at bootup in Fedora rather than a post-install script?

jwrdegoede:
Hi,

On 26-07-17 00:39, Mario Limonciello wrote:
> @seblu <https://github.com/seblu> I generally agree. @jwrdegoede <https://github.com/jwrdegoede> why are the modules built at bootup in Fedora rather than a post-install script?

As I think the original commit message mentions, the modules
are built using a post-install script for:

1) The current running kernel
2) Any kernel installed after the dkms is installed

The problem is:

3) Booting into a different kernel, which was already installed

A typical scenario where we hit this problem is:

a) Run yum update get new kernel
b) Install dkms
c) Reboot (and end up running the new kernel)

That is the scenario which my original commit tries to fix.

Regards,

Hans

yuzaipiaofei:
If someone has more than 100+ dkms modules ,and all the modules need to built during booting time.
It will  need customer to wait a long time to finish the building and Installation work.
And when you update kernels ,the post script will rebuild the dkms for new kernel.

seblu:
Any change to get this merged?

jwrdegoede:
Hi,

On 16-12-17 19:20, Sébastien Luttringer wrote:
> Any change to get this merged?

As mentioned earlier in the discussion, I suggest just removing
these 3 lines added by the commit you want to revert:

if [ -f /lib/systemd/system/systemd-modules-load.service ]; then
	systemctl restart systemd-modules-load.service
fi

Instead of reverting the entire commit. Is there any reason
why just removing those 3 lines would not fix the problems
you are seeing on archlinux?

Regards,

Hans

seblu:
Hi,

I answered you in my July 15th comment; keeping the `modprobe` line is still an issue.

> It's not expected when you build or install modules for another kernel that the running one, to try to load all system modules. Neither when you are updating an already loaded module.

Cheers,

jwrdegoede:
Hi,

On 16-12-17 23:05, Sébastien Luttringer wrote:
> Hi,
> 
> I answered you in my July 15th comment; keeping the |modprobe| line is still an issue.
> 
>     It's not expected when you build or install modules for another kernel that the running one, to try to load all system modules. Neither when you are updating an already loaded module.

You say it is "Not expected" but loading modules as soon as hardware shows up is
done by default, so all modules for which there are modaliases will already
be loaded and this is a no-op. Do you have any bugreports about this bit
causing any actual problems?

Have you tried just removing the 3 lines in the arch dkms package and see what
happens?

Regards,

Hans

seblu:
Hi again,

As far I know, default loading of aliased modules is not done by dkms. Why because I install a virtualbox kernel module, dkms would enter in the game with its own algorithm and start loading my wifi card modules which are **not under its management**?

We don't know what modules have been loaded or unloaded by users and they don't expect that dkms load **unrelated** and **unmanaged** modules. 
Moreover, you can install modules for kernel you don't running on, why again dkms would start loading modules for another kernel that the one requested on the command line?
I will not get into race condition issues between freshly built dkms modules with the same modalias that an already existing one.

So, if this command is a no-op, you wouldn't added it and we can drop it quietly.

I cannot have bug reports about this because I already reverted this commit downstream. But there is no need to push these lines to see what's wrong in it. I'm sorry to not be able to make it enough clear to you.

`dkms install` is about installing files into the right place in the filesystem for a given kernel version.

yuzaipiaofei:
HI @seblu .
I still think the immediate building is necessary ,at least we can see which module can be built successfully for new kernel .and it can  avoid next booting up time too long.
So. i would not like to merge the patch.
But still want to thank you to take the consideration to us.
If you do not mind . i would close the request.

seblu:
Hi @yuzaipiaofei,

Why are you are talking of immediate building? When the modules are built is not the subject of this. We agree that building modules at boot time is not a good idea.
This revert is about **loading** modules which are not built by dkms.

seblu:
Any chances we get this commit reverted?

jwrdegoede:
As I've mentioned already repeatedly please test if just removing these 3 lines (so a partial revert) works:

```
if [ -f /lib/systemd/system/systemd-modules-load.service ]; then
          systemctl restart systemd-modules-load.service
fi 
```

seblu:
I tested it. This still load modules, including modules not managed by DKMS.

jwrdegoede:
"I tested it. This still load modules, including modules not managed by DKMS."

Modules which will all already have been loaded at bootup, unless they are blacklisted, in which case they won't be loaded by the:

find /sys/devices -name modalias -print0 | xargs -0 cat | xargs modprobe -a -b -q

Line since that passes -b to modprobe so it honors the blacklist.

Has anyone actually complained about it loading modules which should not be loaded? The original bug report: https://bugs.archlinux.org/task/54481

Is about the following error: "Job for systemd-modules-load.service failed because the control process exited with error code. See "systemctl status systemd-modules-load.service" and "journalctl -xe" for details."

Removing the 3 lines I suggest to remove will resolve this.

c0d3z3r0:
I am not affected by any issues, since I have applied the Arch Linux patch which reverts this.
However, I saw this PR and want to add my two cents...

I have to agree with seblu. While commit f5bfb12 *maybe* does not introduce any real problems for most users, it should get reverted or at least reworked.

As @seblu already mentioned the line `find ... xargs ... modprobe` does not only load dkms modules but modules completely unrelated to dkms. This is absolutely not logical!

> Modules which will all already have been loaded at bootup, unless they are blacklisted

This is not a valid argument! **Loading** modules in the **install** process is a thing to be discussed... Nevertheless, dkms shouldn't touch/load/dowhatever any modules it has no relation to, out of principle.

Why don't you load (actually dkms-built!) modules in the boot script itself?

rhertzog:
@jwrdegoede See #85 for another case where this change breaks real stuff, this time in the Debian installer while installing packages providing kernel modules via dkms. And no the problem is not only in the systemctl call, it's only in the modprobe line above.

elboulangero:
Lately I was working with DKMS, and I wanted to order DKMS to run before `systemd-modules-load.service`.

So I had a systemd service with `Before=systemd-modules-load.service`, that would run a `Type=oneshot` script, and this script would run dkms.

What happens is that systemd deadlocks, probably because dkms tries to restart systemd-modules-load, from within a oneshot script that I ordered to run *before* systemd-modules-load.

The reason of the failure was not immediately obvious, and I was quite surprised to find that dkms called `systemctl restart` under the hood.

I would have think that the general philosophy with systemd is that the sysadmin takes care of ordering the units as needed, but units themselves shouldn't call on each other, which makes the task of ordering units much more complicated.

So in this case, dkms should NOT restart systemd-modules-load, but instead the sysadmin should take care of ordering dkms before systemd-modules-load, if that's the desired behavior. At least that's my understanding of things.

elboulangero:
Additionally, assume the following scenario (that's also something I'm working on):
- boot an OS image configured with the mesa graphics stack
- if nvidia hardware is detected, change the `glx` alternative from `mesa` to `nvidia` (alternatives are a debian thing)
  - this creates the file `/etc/modules-load.d/nvidia.conf`, that contains the line `nvidia-drm`
- then build nvidia drivers with dkms

In this case, what happens is that dkms restarts systemd-modules-load, which will try to load `nvidia-drm`. However this is not what we want. The nvidia driver can be used only after a reboot, and here this just leads to error messages in the logs. It's not dramatic, but it would be better if it just didn't happen.

adalton:
I have what I think is another rational to apply this change (or at least to have some sort of option to prevent this behavior).

I have a containerized application that includes a kernel module that we use `dkms` to build and load.  Running `dkms` to build and load our specific module has the unwanted side-effect of loading other unwanted modules.  It is very counter-intuitive to call `dkms` to install a module --- by name --- and have it load other unrelated modules.

rhertzog:
Can we get some action on this issue? The issue is crystal clear for me and it breaks use cases where modprobe calls are not desired.

evelikov:
Hey @jwrdegoede , all

Looking at the original commit, it seems like you're interested in having the installed module loaded before the, in dkms, dracut initrd stage kicks in (1).

I'd imagine this would result in the nvidia driver not shipping within the initrd.

Although... the commit message says that the nvidia module is not loaded after a reboot.
This sounds a bit strange and AFAICT this highlights a different problem:
 - (2.1) the depmod stage wasn't properly triggered - hence userspace doesn't know to load the module
 - (2.2) udev is confused failing to load the module, or
 - (2.3) there's a race between loading the module and starting the login manager

If I were to bet, my money goes for a race condition (2.3) which gets hidden by having the nvidia module in initrd (1). An easy way to verify, is to move the `modprobe/systemctl restart` hunk after the `make_initrd`.

If that truly is the issue, one could try the following:
- have a dracut config file or call modprobe [within the Fedora/RHEL package], and
- trigger dracut initrd rebuild within the package

jwrdegoede:
> Looking at the original commit, it seems like you're interested in having the installed module loaded before the, in dkms, dracut make initrd stage kicks in (1).

No, that is not the problem. The problem is the 1-2-3 described in the commit message:
1. User installs a new kernel
2. User does a dnf install nvidia-binary-driver, _while still running the old kernel_, this means that the following conditions are now true:
 a. /etc/kernel/postinst.d/dkms for the new kernel has run before the dkms was present, so the dkms module is not build
  b. The postinstall of the dkms module itself will have built it, but only for the currently running kernel not for the new kernel
 c. a + b together means that the module was NOT build for the new kernel
3. The user reboots into the new kernel, the dkms service now comes to the rescue and builds the module for the new kernel at this point

So there are conditions, which yes users actually hit in practice, where a module does not get build for a new kernel until that new kernel is booted. Note this is the whole reason why we have the dkms boot service, if this were not true we might just as well remove it.

Now the problem is that udev tries to load e.g. the nvidia or the broadcom-wl driver **before** the dkms service builds it at boot. So this means that the user boots into a black screen (nouveau is blacklisted by installing the nvidia dkms module) or without working wifi.

Simply reverting commit f5bfb12fef1fc06e56355cdba500eaa98d4e6aa8 will restore this problem causing a somewhat  severe regression for many users.

May I suggest another solution instead, add a command-line argument to `dkms autoinstall` to either enable or disable this behavior and in case of having it disabled by default, add that command line argument to the dkms invocation in the /lib/systemd/system/dkms.service file so that we do preserve this behavior when building modules at boot. That is a solution which I can get behind.

rhertzog:
I'm sorry but you are fixing this at the wrong place. In your step 2.b, you should build the modules for all available kernel versions... There's even a setting for this in /etc/dkms/framework.conf, it's `$autoinstall_all_kernels`.

> May I suggest another solution instead, add a command-line argument to dkms autoinstall to either enable or disable this behavior and in case of having it disabled by default, add that command line argument to the dkms invocation in the /lib/systemd/system/dkms.service file so that we do preserve this behavior when building modules at boot. That is a solution which I can get behind.

Yes, this is what I was going to suggest, I agree that you might want to use the newly built modules when it gets built on boot... so this behaviour should be disabled by default and enabled only in the boot service.

evelikov:
Thanks for elaborating Hans - as a person who usually doesn't use autoinstall that wasn't obvious from the commit itself.

Some food for thought if I may:
 - If we're talking about "autoinstall" why is the hunk applied in the normal "install" section?
 - Even then: Would it make sense to have "autoinstall --all" triggered as the nvidia-binary-driver gets installed? After all one is installing the binary driver and they would want it to work for _any_ kernel that is present on the system - currently running, newly updated and any others (say LTS perhaps?).

Having the modprobe within dkms seems to be working around the issue, instead of addressing it.

jwrdegoede:
>     * If we're talking about "autoinstall" why is the hunk applied in the normal "install" section?

I honestly do not remember.

> Having the modprobe within dkms seems to be working around the issue, instead of addressing it.

Building the module for all installed kernels might be quite slow. Also there are other possible scenarios where the module-build may simply fail, yet work on reboot, e.g.:

1. User installs dkms but does not have the kernel-devel package for the current kernel
2. User notices builds error and installs kernel-devel package
3. User reboots

Or IOW, there is a reason we have the boot service to build missing modules at boot, so as long as we have / need that service, it should really also load the modules it builds as it runs after the normal auto-loading of modules.

But as said I'm fine with limiting the behavior of doing the modprobe on the modaliases to just the invocation from the boot-service (and then only when it actually had to do any work), which I believe should make everyone happy.

evelikov:
IMHO it's reasonable to shift most of that delay from boot to package update/install time. Thus the package manager/updater can provide for more useful feedback to the user.

Think of frustrated users as their machine is "stuck" at boot, when messages like "dkms: running auto installation service for kernel ..." are hidden by plymouth or otherwise.

Regardless of the above brain-dump, shifting the modprobe to the boot-service mode makes sense.
If people are happy, let's go with that. Thanks

seblu:
> @seblu Are you willing to update your merge request to implement this?

As long dkms loads **only** modules it just installed, have an optional switch would be a plus.

I can update the merge request, but it will be easier to revert this and let @jwrdegoede or someone interested in such feature propose a **correct** implementation which **really** only loads dkms modules. I mean no more find and systemctl reload.

@jwrdegoede Why you don't fix your 2.b to build kernels modules for all the installed kernels?
I implemented such solution in our [pacman hooks](https://git.archlinux.org/svntogit/packages.git/tree/trunk/hook.sh?h=packages/dkms&id=51a9c4a42d7b01414caebee03dca6e6c55f1eb1e) to never build modules at boot.

adrelanos:
What happens in Debian buster when using a `/usr/lib/modules-load.d` drop-in configuration snippet is that the Debian prerm maintainer script runs DKMS which removes that module, when DMKS restarts `systemd-modules-load.service` which causes an unneccessary error message, followed by running the Debian postinst maintainer script which runs DKMS to build module.

Restart of `systemd-modules-load.service` or running `modprobe` is in my opinion out of scope for DKMS. It would be better to leave this running these triggers to the package management system.

If you disagree, please kindly consider making this configurable and add an easy way to drop config snippets [1] so distribution maintainers can easily turn this off.

----

[1] As per `add /etc/dkms/framework.conf.d configuration file drop-in folder`.
https://github.com/dell/dkms/issues/116

scaronni:
> @jwrdegoede Why you don't fix your 2.b to build kernels modules for all the installed kernels?
> I implemented such solution in our [pacman hooks](https://git.archlinux.org/svntogit/packages.git/tree/trunk/hook.sh?h=packages/dkms&id=51a9c4a42d7b01414caebee03dca6e6c55f1eb1e) to never build modules at boot.

Just a note on this, the kernel + module packages are protected during an update (kernels are installed in parallel) but the additional kernel packages are updated on the fly, so you can not build for all kernels unless you also keep all devel packages around which is not how the distribution works.

seblu:
I see you issued a new release @GoPerry. Nice !
Any chance to take a look at this ?

scaronni:
Merged and disabled by default, it will be enabled by default in the Fedora / RHEL / CentOS package with an appropriate configuration file. Thanks @evelikov.

PR #34, opened 2017-09-26
dkms build: Don't ignore -c if module is already built

martingalvan:
Currently, dkms build will always look for dkms.conf in
`$dkms_tree/<module>/<version>/source`, even if a custom one was specified
hrough -c. This is problematic for cases where dkms.conf is located in some
subdir of the module tree.

This patch changes is_module_built so that it'll first check whether $config
is set.


yuzaipiaofei:
looks good to me.merged.
Thanks. should be more convenient to use specific dkms.conf .

PR #39, opened 2017-11-27
Compressed module support

### LINKED ISSUE #38: Compressed modules (*.ko.gz)
I installed a patch for alx.ko module and i noticed the original module is alx.ko.gz, but DKMS puts alx.ko into /lib/modules/4.13.12-1-ARCH/kernel/drivers/net/ethernet/atheros/alx/ thus kernel is loading the wrong module!

How can i make DKMS create and replace the alx.ko.gz module?

I'm running Arch Linux.

d3matt:
Should resolve #38

### DIFF
--- a/dkms
+++ b/dkms
@@ -174,16 +174,11 @@ set_module_suffix()
 {
     # $1 = the kernel to base the module_suffix on
     kernel_test="${1-:$(uname -r)}"
-    module_suffix=".ko"
-    [[ $(VER $kernel_test) < $(VER 2.5) ]] && module_suffix=".o"
-    grep -q '\.gz:' /lib/modules/$kernel_test/modules.dep 2>/dev/null && gz_modules=1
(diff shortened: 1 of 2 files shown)


d3matt:
Realized there are a couple more cases it needs to handle. Will update PR tommorow.

scaronni:
Hello @d3matt, is this finished? Is your latest commit referred to the above comment?

Thanks.

d3matt:
I think there's still some interaction between weak-modules and dkms for the mixed case (kernels with compressed modules and kernels with uncompressed modules installed), but I haven't been able to pinpoint what's going on there. My gut feeling is that weak-modules needs an update as well, but I don't have evidence to point in that direction. This PR is probably ready to be merged, but someone else might want to play with it as well.

scaronni:
Thanks, will try it soon.

scaronni:
Works perfectly. Sorry for the delay.

Issue #25, opened 2017-06-13
using rpm_safe_upgrade correctly?


bitness:
Hello,

I have an RPM that uses dkms with --rpm_safe_upgrade in its %post and %preun sections, and it uses them essentially the same way as in dkms's sample.spec:

    %post
    dkms add -m modname -v %{modver} --rpm_safe_upgrade
    dkms build -m modname -v %{modver}
    dkms install -m modname -v %{modver}
    
    %preun
    dkms remove -m modname -v %{modver} --all --rpm_safe_upgrade

As far as I can tell, though, this doesn't work when upgrading from one release of the package to the next, without the module version changing. `dkms add` with `--rpm_safe_upgrade` exits out with a complaint that the module version is already added, but it exits before writing the rpm_safe_upgrade lock file. When `dkms remove` is called later in the%preun section, it sees no lock file and so uninstalls the module.

Shouldn't add_module() write the rpm_safe_upgrade lock file _before_ exiting due to an existing module with the same version? I hacked my /usr/sbin/dkms to do this and it all worked as expected.

Thanks!
-lars

yuzaipiaofei:
--rpm_safe-upgrade prevents the uninstallation of the driver in case a package with the same version-release is reinstalled.

yuzaipiaofei:
You can refer to dkms package policy.
[](agile.us.dell.com/Agile)

yuzaipiaofei:
If you do not have any more questions about this .i would like to close this case in this week.

bitness:
Hi,

Sorry, I don't see any content at http://agile.us.dell.com/Agile.

The dkms man page says this about rpm_safe_upgrade:

> Its use makes sure that when upgrading between different releases of an RPM for the same <module-version> , DKMS does not do anything dumb

So it's not just intended for uninstallation protection during reinstallation of the same version-release, it's specifically called out as the right thing to do when the module-version hasn't changed, but the release has.

[...]

PR #44, opened 2018-01-24
fix add_module rpm_safe_upgrade handling

scaronni:
This change makes add_module() write out the rpm_safe_upgrade lock
_before_ checking to see if this module-version has already been added
when --rpm_safe_upgrade is specified.

Before, if there was a module-version match, no rpm_safe_upgrade lock
file would get written, and the old package's %preun script's `dkms
remove` call would proceed when it should have stopped.

### DIFF
--- a/dkms
+++ b/dkms
@@ -1038,6 +1038,14 @@ add_module()
     # Check that we have all the arguments
     check_module_args add
 
+    # Do stuff for --rpm_safe_upgrade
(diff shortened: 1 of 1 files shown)


scaronni:
Created PR for proper testing of this: https://github.com/dell/dkms/issues/25

scaronni:
Shall we merge this?

superm1:
> Shall we merge this?

I think so.

PR #45, opened 2018-01-24
Fix version and date in manpage, be consistent with variable replacement

scaronni:
And make sure the date is pretty printed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-RELEASE_DATE := "30-Nov-2017"
+RELEASE_DATE := "30 November 2017"
 RELEASE_MAJOR := 2
 RELEASE_MINOR := 5
 RELEASE_NAME := dkms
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -32,17 +32,17 @@ clean:
...
 	install -p -m 0755 dkms_autoinstaller $(LIBDIR)
 	install -p -m 0644 dkms_framework.conf $(ETC)/framework.conf
 	install -p -m 0644 dkms_dbversion $(VAR)
 	install -p -m 0644 dkms.bash-completion $(BASHDIR)/dkms
-	# install compressed manpage with proper timestamp and permissions
-	gzip -c -9 dkms.8 > $(MAN)/dkms.8.gz
-	chmod 0644 $(MAN)/dkms.8.gz
+	install -p -m 0644 dkms.8 $(MAN)
+	# add versioning
+	sed -i -e 's/RELEASE_VERSION/$(RELEASE_VERSION)/' -e 's/RELEASE_STRING/$(RELEASE_STRING)/' -e 's/RELEASE_DATE/$(RELEASE_DATE)/' $(SBIN)/dkms $(MAN)/dkms.8
```
superm1: Would you mind to use some marking characters like ## around the match string?  This helps to prevent inadvertently causing weird behavior down the road when someone forgets this sed statement is part of `make install`.

Makefile
```diff
@@ -1,4 +1,4 @@
-RELEASE_DATE := "30-Nov-2017"
+RELEASE_DATE := "30 November 2017"
```
superm1: I would argue just remove release date.  it doesn't give a lot of value to anything these days.  It's just one more thing to forget to do when tagging a new release.

yuzaipiaofei: The Release date will be changed when the new version released. 
So the value should be limited. but  it is ok for me to change here.

PR #50, opened 2018-04-04
deleting link to the dell's white-paper (not available anymore)

alexander-naumov:


superm1:
Thanks, forgot those were casualties with the move.

Issue #74, opened 2019-01-18
`dkms mkdeb` does not include binaries


npf:
Hi,
`dkms mkdeb` seems to not include the .ko module files (binaries).
Either this is a bug, or preferably this is just the good way to go, but usage and man page should be fixed so that --source-only is de facto, an --binaries-only is not an option.

Also note that if `mkdeb` could actually produce a deb with binaries inside, it should not be `_all.deb`, but an architecture specific deb package (e.g. `amd64.deb`).

My proposal would be to take the approach chosen in Debian's dkms version (fork):
- have a `mkdeb ` command to create the dkms source package, `all.deb`
- have a `mkbmdeb` command to create the dkms binary modules package (arch specific package)
This means merging Debian's patches for `mkbmdeb`, and fixing `mkdeb` usage.
See: https://salsa.debian.org/debian/dkms/blob/master/debian/patches/0004-mkbmdeb-support-for-lean-binary-package-with-only-th.patch

This is something discussed in Debian bug tracking system here:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=832558#100

GoPerry:
patches merge,close the issue.
Thanks for the codes contribution.
please feel free to reopen the issue if you still have any questions for this.

PR #75, opened 2019-01-19
Add mkbmdeb command, rework mkdsc, mkdeb commands

npf:
The mkbmdeb command is shipped with Debian's dkms version (fork). This PR imports the related patch + fixes it.

- mkbmdeb creates a Debian package with the precompiled binary driver only (.ko module files).

- mkdeb creates a Debian package providing the dkms sources of the driver, no binaries. It is source-only de facto. same for mkdsc → this PR drops the --binaries-only and --source-only for mkdeb and mkdsc.

This PR is the result of a discussion in Debian bug tracking system for dkms:
 https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=832558#100

It fixes issue #74 .


yuzaipiaofei:
LGTM ,thanks for your patches.

PR #77, opened 2019-04-17
Revert "FIX dkms built issues for rhel kernel update ,change prerm sc…

sil2100:
…ript to fix."

This reverts commit 0c19129b5d1f8e03498f6f2455ad9f7e14e9e606.

Creating this PR also for discussion, as we think the current behavior is wrong and not fitting the design defined in the manpage.

Per man dkms, the `uninstall` command should leave the uninstalled modules in the 'built' state and only removed after running `remove`. After this commit it seems that dkms uninstall started to remove the modules from the built tree, making them no longer stay in the 'built' state (they're not reported as such anymore).

This caused a regression in Ubuntu 19.04 where some of our other packages expected things to work as per the manpage. Quoting:
"Uninstalls an installed module/module-version combo from the kernel/arch passed in the -k option, or the current kernel if the -k option was not passed upon.  After uninstall completion, the driver will be left in the built state.
To completely remove a driver, the remove action should be utilized."

Without this change reverted, this does not seem to be the case anymore. After the revert, `dkms uninstall` seems to work as expected (and as it previously did). What was the rationale for the original change? What use-case was it fixing? The change description wasn't very descriptive.

Thank you!

### DIFF
--- a/dkms
+++ b/dkms
@@ -1782,9 +1782,7 @@ do_uninstall()
             fi
         fi
     done
-    rm -rf "$dkms_tree/$module/kernel-$1-$2"
-    #Remove modules from dkms built tree
-    rm -rf "$dkms_tree/$module/$module_version/$1"
+    rm -f "$dkms_tree/$module/kernel-$1-$2"
     else
         echo $""
         echo $"Status: This module version was INACTIVE for this kernel."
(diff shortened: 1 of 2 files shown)


superm1:
@yuzaipiaofei can you please review this?

yuzaipiaofei:
LGTM

matneh:
Holy cow, I just ran into this exact issue, I can't believe it's been around for 2 years!  Thanks for the fix!

PR #87, opened 2019-07-05
implement a simple module signing mechanism

xuzhen:
usage:
add ```SIGN_TOOL=/path/to/signtool.sh``` into dkms.conf or /etc/dkms/modulename.conf
the signtool.sh is something like
```
#!/bin/sh
/usr/src/linux-$1/scripts/sign-file sha512 /path/to/signing_key.pem /path/to/signing_key.x509 "$2"
```


vvvrrooomm:
@xuzhen while trying to use this I found the configuration difficult:
With more than one dkms-module each hast to be configured to use the SIGN_TOOL. 
What do you think, should it be moved to framework.conf? I can supply a PR if this is agreed upon

xuzhen:
@vvvrrooomm I agree that there needs to be a global SIGN_TOOL for all modules. And I prefer to use a global /etc/dkms/dkms.conf. It also overrides the dkms.conf that comes with each module, but has a lower priority than /etc/dkms/<module>.conf

PR #91, opened 2019-07-30
fix: OBSOLETE_BY be blocked by --force (Closes #89)

### LINKED ISSUE #89: priority of OBSOLETE_BY should higher than "force".
there's a scenario that a dkms came form older kernel which be used to override new kernel module for regression.
for example:
1. kernel regression comes from kernel 5.0 on A driver.
2. user take dkms  A from kernel 4.15 to make the device work before the regression be fixed.
3. user install dkms A by --force, so that dkms A can be installed even A driver version higher than DKMS A.
4. the regression is fixed in kernel 5.2.
5. add OBSOLETE_BY="5.2" to dkms A, so that user can still use DKMS A on kernel older than 5.2

But, if the priority of "force" higher than OBSOLETE_BY, then the DKMS A will never be retired.

alex-tu-cc:
OBSOLETE_BY should higher priority than --force.
the idea is same as BUILD_EXCLUSIVE_KERNEL not be blocked by --force.

### DIFF
--- a/dkms
+++ b/dkms
@@ -776,7 +776,7 @@ check_version_sanity()
         local -a my=(${1//-/ })
         local obsolete=0
         if [[ ${obs} && ${my} ]]; then
-            if [[ $(VER ${obs}) == $(VER ${my}) && ! $force ]]; then
+            if [[ $(VER ${obs}) == $(VER ${my}) ]]; then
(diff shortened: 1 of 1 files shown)


yuzaipiaofei:
LGTM

PR #93, opened 2019-08-01
Fix kernelsourcedir usage

Glandos:
`--kernelsourcedir` was completely unused, since `set_kernel_source_dir` and `_get_kernel_source_dir` overwrote it.

This is a quick try. I don't know if it covers all cases.


yuzaipiaofei:
LGTM

PR #96, opened 2019-08-07
man: fix manpage-has-errors-from-man lintian warnings

vicamo:
Yet another error found in https://salsa.debian.org/vicamo-guest/dkms/-/jobs/257553.

Signed-off-by: You-Sheng Yang <vicamo@gmail.com>


vicamo:
[Then](https://salsa.debian.org/vicamo-guest/dkms/-/jobs/257640) it looks like there is no more common lintian warnings/errors but debian specific ones. Thank you.

Issue #42, opened 2018-01-01
Is there a way to skip version check from dkms.conf?


bp0:
Is there some way to make auto install always --force from dkms.conf? Or perhaps, some way to skip the version check? If the module is using the git hash as its module version, then it is not comparable in any regular way.
I tried to use PRE_INSTALL="true", as the man page said that PRE_INSTALL could be used for a custom version check, and that did not work. Also, I tried PRE_INSTALL="dkms install --force -m ..." in an attempt to just hack around the version check, but it didn't work either.

This is the attempted use of DKMS that fails when the kernel package is updated, for reference: https://github.com/groeck/it87/commit/6c1e48b4f16b65b6e2ff95d950b250581b0cbe91

yuzaipiaofei:
The  --force option to be used when the version checking failed.
I do not think we can always need to install packages with force .
@bp0

bp0:
DKMS' version check appears to always fail if the module doesn't provide a module version. For it87, the solution was simply to add a MODULE_VERSION() to the source (https://github.com/groeck/it87/commit/5edd7bb6f908b66b3a2a0d0a7ec9e04d723147e5). To me it still seems like there should be a way to skip the version check, or override it, via dkms.conf.

yuzaipiaofei:
@bp0 ,I think the module missing is the problem of the module, it is not the right way to ignore the normal check process for dkms .
The issue will  be close if you have no further questions about this.

alex-tu-cc:
a related issue that becuase of that weird version comparing rule to compare the git hash that some released dkms add "--force" by default. Then obsoleted_by parameter in dkms.conf will be always be overrided. So, I commited a patch there : https://github.com/dell/dkms/issues/89 , to make obsoleted_by has higher priority than "--force"

But to think a bit more, it might impact people who really want to use "--force" to override everything blindly. So, I'm thinking to remove the part of comparing git hash which looks not really make sense. 

So, the proposal will be: if the module version is not defined, then install that dkms without comparing git hash. Say, people must define a module version if they would like to utilize that version comparing. After that, we can carry back "--force" to own the highest priority to override everything blindly and obsoleted_by will not be impacted anymore.

How do you think?

alex-tu-cc:
after discussing with some other engineers, it seems better to preserve current version checking logic because we don't know who might count on it to build their logic on their system.

[...]

PR #98, opened 2019-09-25
Force version override by file in /usr/share/dkms/modules_to_force_install/

alex-tu-cc:
for https://github.com/dell/dkms/issues/42, I reverted 
546195f and add a new parameter, because there may some people counting on '--force' as highest priority for installation.

so let's list scenarios be covered here:
1. user needs to force install the dkms without any checking. 
 - this is a way which already exists, not be changed by this patch.
 - by '--force' or write dkms name in files under /usr/share/dkms/modules_to_force_install/
 - in this case, obsolete_by will not work, because of '--force' is the highest priority which behaves same as the logic before comment 546195f.

2. user need to force install the dkms, but also would like to retire that dkms once kernel fixes the issue.
 - this way is added by this patch.
 - by '--force-version-override' or write {dkms name}_version-override in  files under /usr/share/dkms/modules_to_force_install/
 - then obsolete_by in dkms.conf will get higher priority.
 - log for a test installation with version override only
   - to simulate the case that fix not in kernel so user need a dkms
   - http://paste.ubuntu.com/p/9bw9r7wrbZ/
 - log for a test installation with version override and obsoleted_by
   - to simulate the case that fix already in kernel so user wants an online update retiring dkms.
   - http://paste.ubuntu.com/p/nrGtygyF3X/

### DIFF
--- a/dkms
+++ b/dkms
@@ -784,7 +784,7 @@ check_version_sanity()
         local -a my=(${1//-/ })
         local obsolete=0
         if [[ ${obs} && ${my} ]]; then
-            if [[ $(VER ${obs}) == $(VER ${my}) ]]; then
+            if [[ $(VER ${obs}) == $(VER ${my}) && ! $force ]]; then
                 # They get obsoleted possibly in this kernel release
                 if [[ ! ${obs[1]} ]]; then
                     # They were obsoleted in this upstream kernel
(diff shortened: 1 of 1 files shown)


yuzaipiaofei:
Alex,thanks for your patch .
LGTM

bp0:
@alex-tu-cc, thanks for working on this.

PR #99, opened 2019-10-02
Switch to shortopt equivalents for some commands

E5ten:
This increases compatibility with various implementations of these commands like busybox and toybox.


yuzaipiaofei:
LGTM

PR #104, opened 2019-10-25
Add support for UEFI Secure Boot validation toggling through shim

LocutusOfBorg:
This allows one to install third-party drivers and still have their system
work (albeit without full Secure Boot validation), automatizing the step
of disabling validation in shim.

Users still have the possibility to notice this and skip the step, thus
not disabling Secure Boot (but then, they will not be able to load the
dkms drivers that were added).

### DIFF
--- a/dkms
+++ b/dkms
@@ -1416,6 +1416,7 @@ clean_build()
 
 sign_build()
 {
+    [[ -x "$(command -v kmodsign)" && -d "/var/lib/shim-signed/mok/" ]] || return
     local base_dir="$dkms_tree/$module/$module_version/$kernelver/$arch"
     if type update-secureboot-policy >/dev/null 2>&1; then
         echo $"Signing module:"
--- a/dkms
+++ b/dkms
@@ -1414,10 +1414,29 @@ clean_build()
     rm -rf "$dkms_tree/$module/$module_version/build"
 }
 
+sign_build()
+{
+    local base_dir="$dkms_tree/$module/$module_version/$kernelver/$arch"
+    if type update-secureboot-policy >/dev/null 2>&1; then
+        echo $"Signing module:"
+        SHIM_NOTRIGGER=y update-secureboot-policy --new-key
+        for ko in `find "$base_dir/module/" -name "*.ko" -print`;
+        do
+            echo " - $ko"
+            kmodsign sha512 \
+                /var/lib/shim-signed/mok/MOK.priv \
+                /var/lib/shim-signed/mok/MOK.der \
+                "$ko"
+        done
+        update-secureboot-policy --enroll-key
+    fi
+}
+
 build_module()
 {
     prepare_build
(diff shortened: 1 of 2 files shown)


dkms
```diff
@@ -1414,10 +1414,29 @@ clean_build()
...
+sign_build()
+{
+    local base_dir="$dkms_tree/$module/$module_version/$kernelver/$arch"
+    if type update-secureboot-policy >/dev/null 2>&1; then
+        echo $"Signing module:"
+        SHIM_NOTRIGGER=y update-secureboot-policy --new-key
+        for ko in `find "$base_dir/module/" -name "*.ko" -print`;
+        do
+            echo " - $ko"
+            kmodsign sha512 \
```
superm1: I think this makes an implicit assumption that `kmodsign` is installed already.  As it's distributed in `sbsigntool`, someone might have removed it.  can you add a check that it exists?

dkms
```diff
@@ -1414,10 +1414,29 @@ clean_build()
...
+                /var/lib/shim-signed/mok/MOK.priv \
```
superm1: Similarly this makes an assumption about the directory structure being used.  I think that needs to be looked up dynamically or at least test that this structure already exists.

LocutusOfBorg:
@superm1 I did a fix for this, let me know if you like it!

dkms
LocutusOfBorg: The idea is to see if other distro starts using this as Ubuntu did, so in this case we can extend (or I'm sure they will do it)...

dkms
```diff
@@ -1414,10 +1414,30 @@ clean_build()
     rm -rf "$dkms_tree/$module/$module_version/build"
 }
 
+sign_build()
+{
+    [[ -x "$(command -v kmodsign)" && -d "/var/lib/shim-signed/mok/" ]] || return 1
```
superm1: returning 1, won't this cause failures for anyone missing these tools or directory structure?

LocutusOfBorg: ok you are right, better just "return", but in the future we should really add a way to tell the user that something is missing...

superm1:
Much better, thanks!

LocutusOfBorg:
thanks to you for merging!

Tachi107:
Hey @LocutusOfBorg, sorry for the ping. I've been looking into using this automatic signing feature of dkms on Debian, but I'm unable to find `kmodsign`. After a bit of research, it seems that this is something Ubuntu-specific. Am I right?

Do you think it would be possible to make dkms use [`scripts/sign-file`](https://www.kernel.org/doc/html/latest/admin-guide/module-signing.html#manually-signing-modules) instead, so that other distros can benefit from this nice feature as well?

Thanks :)

LocutusOfBorg:
yay, I think its possible, even though I didn't implement the code of this pull request :)

PR #118, opened 2020-02-09
Add script for newer Fedora to run existing postinst/prerm scripts

mhjacks:
Pursuant to Fedora bug https://bugzilla.redhat.com/show_bug.cgi?id=1753044, and openzfs issue https://github.com/zfsonlinux/zfs/issues/9868.

The issue is that there is surprising behavior to users expecting that ZFS (and other DKMS-built modules in Fedora) are not auto-built for new kernels by the out-of-the-box configuration of DKMS.  This has nothing to do with ZFS per se; it is a change in how Fedora invokes kernel installation hooks.  DKMS in its present state depends on an older behavior.

Fedora has moved away from effectively requiring the grubby package to run the postinst/prerm scripts in newer Fedoras (30+).  However, the grubby package is still available and usable.

My proposal is to add a new script in the new /etc/kernel/install.d directory, which selectively calls either the existing postinst or prerm script, which would stay in their current locations.  F30+ DKMS packages in Fedora can add the new install.d script.  My hope is that the other OS's that DKMS supports can simply ignore the new script and not have to make changes.

This is what I suspect to be the least disruptive way of handling the problem.  If this PR is accepted I intend to work with the Feodra packaging of DKMS to bring it to present and future supported Fedora releases.  If I am wrong in any of my assumptions, I'm happy to refine my submission.

### DIFF
--- a/kernel_install.d_dkms
+++ b/kernel_install.d_dkms
@@ -1,0 +1,9 @@
+#!/usr/bin/bash
+
+if [[ "$1" == "add" ]]; then
+	/etc/kernel/postinst.d/dkms $2
+fi
+
+if [[ "$1" == "remove" ]]; then
+	/etc/kernel/prerm.d/dkms $2
+fi


gregory-lee-bartholomew:
I've been using the proposed script on Fedora since September of 2019 without issue. However, I also remove the grubby,grub2-*, and os-prober packages and run either the systemd-boot boot loader or syslinux boot loader depending on whether the system is UEFI or BIOS. I have not tested what happens if this script is used with grubby and grub2. I assume it would cause the DKMS scripts to get called twice. Will DKMS see that the modules are already built and immediately exit with success if they are called a second time?

gregory-lee-bartholomew:
BTW, please also note that the naming of the script is significant. It needs to be `/usr/lib/kernel/install.d/40-dkms.install` so that it will run prior to `/usr/lib/kernel/install.d/50-dracut.install`.

mhjacks:
My intention was to handle the numbering in the fedora packaging (e.g. install the new file as 40-dkms.install; that seems to be a fedora-specific thing so I followed the naming convention of the other hook scripts (granted they don't use numbering hints).  I'm willing to change that if the maintainers wish.

I think you're right to be concerned that the presence of grubby could cause duplicate module builds; even if that doesn't actually hurt anything, it would mean the user has to spend longer than they should in the RPM transaction.  I ran the current script as proposed in place with grubby installed, and it appeared to me that the module was only built once (by the 40-dkms.install script).  It seems the autoinstall logic is clever enough to realize that the module matching the version has been built and exit without rebuilding it. (the script runs `dkms autoinstall`)  So I think we'll be OK there.

PR #131, opened 2020-06-17
dkms: use /etc/dkms/no-autoinstall as flag file to disable autoinstall

LocutusOfBorg:
intended use: autopkgtests should not fail while installing *-dkms
 or linux-headers-* without detailed error reporting
 instead they can run the invidiual steps with verbose error reporting


LocutusOfBorg:
hello, ping?

LocutusOfBorg:
thanks!

PR #132, opened 2020-06-17
Makefile: also install in /etc the kernel_install.d_dkms and sign_hel…

LocutusOfBorg:
…per.sh scripts


LocutusOfBorg:
TODO: should we also install them with debian and redhat targets?

scaronni:
> TODO: should we also install them with debian and redhat targets?

Yes please. Thanks.

LocutusOfBorg:
Rebased and pushed!

LocutusOfBorg:
thanks!

PR #136, opened 2020-07-31
Do not require root access for the build command

### LINKED ISSUE #134: dkms build with --dkmstree/--sourcetree does not work anymore without root access
Running this command without root access now fails for me:

```
$ dkms build --dkmstree "$(pwd)" --sourcetree "$(pwd)/NVIDIA-Linux-x86_64-450.57-no-compat32" -m nvidia/450.57 -k "$(</usr/src/linux/version)"
Error! You must be root to use this command.
$ dkms --version
dkms:2.8
$ ls NVIDIA-Linux-x86_64-450.57-no-compat32/nvidia-450.57/
common  conftest.sh  dkms.conf  Kbuild  Makefile  nvidia  nvidia-drm  nvidia-modeset  nvidia-uvm
```

Using  dkms 2.8.3. This seems to be caused by commit https://github.com/dell/dkms/commit/bd13868dcb1d9d6b89cc60edcd5a7a022bce516e.

Building modules without root access was perfectly fine to up until version 2.8.1 by using the `--dkmstree` and `--sourcetree` options. The dkms [man page](https://github.com/dell/dkms/blob/v2.8.3/dkms.8#L541-L543) says that the `--sourcetree` option is "_Useful for systems that you may not have root access, but would still like to be able to build DKMS packages_", clearly stating that it should be possible to build modules without root access. This is useful, for example, for distro packagers.

Any chance that it could be possible to build modules without root access again when using `--dkmstree`/`--sourcetree`?

dbermond:
Commit bd13868dcb1d9d6b89cc60edcd5a7a022bce516e broke the build command when running without root access. Building as a non-root user is useful in some cases and is described in the man page as a supported feature.

See the discussion in #134

Fixes #134

### DIFF
--- a/dkms
+++ b/dkms
@@ -3886,7 +3886,8 @@ for action_to_run in $action; do
         check_root && make_driver_disk
         ;;
     build)
-        check_root && check_all_is_banned "build" && build_modules
+        [ ! -w "$dkms_tree" ] && die 1 $"No write access to create a DKMS tree at ${dkms_tree}"
+        check_all_is_banned "build" && build_modules
         ;;
     add)
         check_root && check_all_is_banned "add" && add_module


evelikov:
Can we please update the man page? Currently there is no reference between the build and sourcetree sections or vice-versa.

PR #139, opened 2020-09-04
Compatibility with ALT Linux

vt-alt:
Mainly:
* `/etc/sysconfig/kernel` is directory.
* Do not `rmdir` module installation directory if it's belonging to rpm package. (After second thought I think this patch is not so important, but it does not hurt either.)
* We use `dkms_test-1.0` for dkms CI testing, make its `dkms.conf` suitable for it, removing `MAKE` & `CLEAN`options with obsolete build syntax. Add `AUTOINSTALL` which is (I think is) useful for example and testing.

### DIFF
--- a/dkms
+++ b/dkms
@@ -1823,7 +1823,7 @@ do_uninstall()
         while [ "${dir_to_remove}" != "${dir_to_remove#/}" ]; do
             dir_to_remove="${dir_to_remove#/}"
         done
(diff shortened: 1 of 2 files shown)


scaronni:
`/etc/sysconfig/kernel` is actually a file also on CentOS/RHEL/Fedora. Checking if it creates any compatibility issue there where the check has always been missed.

PR #144, opened 2020-11-11
Add support for zstd compressed modules

Sporif:
Since zstd support was added to kmod ([3821e197](https://git.kernel.org/pub/scm/utils/kernel/kmod/kmod.git/commit/?id=3821e1971ec53fabbbb9f2679ea988ee12db61c8)) and can be added to the kernel with a simple [patch](https://aur.archlinux.org/cgit/aur.git/tree/0009-add-zstd-compressed-modules.patch?h=linux-nc&id=4c775afe4701a6896689230bac025c21dd6d00bf).


scaronni:
@Sporif Is the zstd patch mainline in the kernel already?

Sporif:
@scaronni No, not as far as I know. Is that a requirement for this being merged?

scaronni:
No, just asking as I'm not in favour of keeping "dead" code around. There's already enough EOL distributions still supported by DKMS that would be nice to remove. I can work on the CentOS/RHEL/Fedora part, but not more unfortunately.

PR #156, opened 2021-09-23
kernel_prerm.d: fix for new dkms status format

iomartin:
Commit f83b758b6fb8ca67b1ab65df9e3d2a1e994eb483 ("Print 'module/version'
and not 'module, version' when doing dkms status.") changed the output
format of dkms status command.

This broke the prerm.d script.

Fix to use the new format


scaronni:
Thanks!

PR #158, opened 2021-10-02
Drop distro-specific packaging

scaronni:
This branch removes the distribution specific packaging, and leaves only the option for installing from the source using a Makefile. I think the packaging should not be done by the project but instead from the various distributions.

- Debian stuff is severely out of date compared to upstream, and it will never be if Debian updates again for the next version: https://salsa.debian.org/debian/dkms
- SPEC file can not be used as is on Fedora/Redhat and recent SUSE versions as it does not work with custom information in `.rpmmacros` and lots of other issues.
- SPEC file covers lots of outdated distributions but not recent ones, project will be never be able to keep up to date with all distributions.
- Most distributions (all?) already provide packages in their own packaging guidelines (Debian, Ubuntu, Fedora, CentOS/Redhat, Arch, Mageia, etc.).
- People open bugs saying that the packaging part does not work, no patches are provided for improving the situation by people knowledgeable on the distribution reported.
- There are no contributoris with enough distribution specific knowledge.

At this point I hardly believe that the distro-specific code is actually usable in the current form.


evelikov:
Fwiw I could not agree more.

Distributions know the quirks and specifics for their platforms. If they request that some helper is added to dkms - then sure. Say for example, they want a stable format for the module name/version/status/etc. But implementing deb/rpm packaging seems like a serious overkill.

Note that with this, a bunch of targets are unused and can be removed:
 - debian packages - 200 loc
 - rpm/kmp - 160 loc

Personally I'm inclined that a few more targets can go:
 - driverdisk - floppy, rhel3 support really? - 500 loc
 - initrd and sysconfig - covers a very limited set initrd generators - 300 loc
 - depmod - very time consuming to run, so distros might want to run it once - 50 loc
 - supporting linux kernel earlier than 3.0 - 50

So overall, apart from the 2 kloc removed here, one can drop another 1.2 kloc from dkms alone - that's 1/3 of dkms.

Keep in mind that there are personal preferences and I do not speak on behalf of any distribution.

HTH o/

PR #163, opened 2021-10-09
Makefile fixes - mkdir + install -> install -D, use git archive ...

evelikov:
Few trivial fixes, making the file a bit cleaner and easier to read.


scaronni:
I only reverted the gzipping of the man pages. Distribution tools already take care of compressed and non compressed cases, but for cases where people just install on a system with `make install` (like i've seen in some other bug reports) then they would miss the compression.

evelikov:
Man works with uncompressed files, doesn't it? If so what's the problem with leaving them uncompressed for `make install` users?

Not a big deal either way, just a bit strange.

PR #165, opened 2021-10-10
Allow removing multiple kern/arch in different state

evelikov:
Earlier I had 2 different modules in different states across 3 different
kernels. So instinctively I've used `dkms remove module/ver --all` only to
be greeted by a failure.

If people want to remove/unbuild/uninstall module across different kern/arch,
they do not care if one of those already is removed/unbuild/uninstalled.

Convert the "or_die" instances to "maybe", which also makes it similar
to the additive (add/build/install) paths.

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -1843,18 +1843,25 @@ module_is_added_or_die()
         $"The module/version combo: $module-$module_version is not located in the DKMS tree."
 }
 
-module_is_built_or_die()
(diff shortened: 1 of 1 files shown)


evelikov:
Note this actually fixes a bug, where you cannot `dkms remove`, if the module is not in `build` state.

evelikov:
Rebased

Issue #162, opened 2021-10-08
Rework module signing


scaronni:
Configuration should be as follows:
- Create private key and public key in `/etc/dkms`
  - Location of the keys should be configurable, for users who want to keep their distribution defaults
- Enable configuration to use the keys in `/etc/dkms/framework.conf`
- Modules are signed at every rebuild.

Configuration:
- Entries in `/etc/dkms/framework.conf` for keys
- File `/etc/dkms/mok.key` points to the Machine Owner Key private key
- File `/etc/dkms/mok.der` points to the Machine Owner Key certificate in DER format

Main command:
- No more old specifics Ubuntu commands in the `sign_build` function
  - Detect the distribution and update the `sign_helper.sh` script with the correct distribution path for `sign-file`
- Merge of the `sign_helper.sh` script in the `sign_build` function
- Drop `sign_tool` variable and documentaton

Docs:
- Add man page entry on the new configuration.
- Update README.md with information and screenshots for MOK enrollment

Later step:
- Drop the `sign_helper.sh` entry.

scaronni:
@Tachi107 I've updated the description for the certificate location, this should cover your requirements.

scaronni:
HI @Tachi107, please have a look at: https://github.com/dell/dkms/tree/rework_signing

It's a work in progress, but I think covers all use cases. You can have your DKSM keys managed the "Debian way" if you want to but at the same time covers any other use cases. The complexity is also pretty low.

scaronni:
Sample output:
```
# dkms build nvidia/515.43.04
Sign command: /lib/modules/5.17.7-300.fc36.x86_64/build/scripts/sign-file
Signing key: /etc/dkms/mok.key
Public certificate (MOK): /etc/dkms/mok.pub
Creating symlink /var/lib/dkms/nvidia/515.43.04/source -> /usr/src/nvidia-515.43.04

Building module:
Cleaning build area...
'make' -j2 module SYSSRC=/lib/modules/5.17.7-300.fc36.x86_64/build IGNORE_XEN_PRESENCE=1 IGNORE_PREEMPT_RT_PRESENCE=1 IGNORE_CC_MISMATCH=1...........................
Signing module /var/lib/dkms/nvidia/515.43.04/build/nvidia.ko
Signing module /var/lib/dkms/nvidia/515.43.04/build/nvidia-modeset.ko
Signing module /var/lib/dkms/nvidia/515.43.04/build/nvidia-drm.ko
[...]

PR #171, opened 2021-10-30
Random selection of fixes and cleanups

evelikov:
Grouping the lot in a single MR since they fit no particular topic:

 - remove multi-action machinery - manual says one action, code "supports" multiple
 - move input arguments validation earlier
 - use local -r variables - makes the code a bit easier on the eyes
 - update tar invocation - drop cd push/pop, auto-detection magic
 - always check for write permissions in make_tarball
 - add some early return/break statements - try to tame the crazy indentation, without rewriting the world

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -1263,12 +1263,6 @@ install_modules() {
     done
 }
 
-check_module_exists() {
-    is_module_added "$module" "$module_version" && return
-    die 2 $"DKMS tree does not contain: $module-$module_version" \
-        $"Build cannot continue without the proper tree."
-}
-
 possible_dest_module_locations()
 {
     # $1 = count
@@ -1637,8 +1631,6 @@ show_status()
 
 make_tarball()
 {
-    make_common_test
-
     # Read the conf file
     read_conf_or_die "$kernelver" "$arch"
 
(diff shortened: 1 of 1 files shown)


evelikov:
Seems like we're trying to sign the modules twice atm - once in `do_build` and second time in `sign_build`. The latter cannot work in most cases, since it's looking for uncompressed kernel modules - where most distros compress them.

I'm inclined that we should nuke `sign_build` and inline `prepare_build` `do_build` and `clean_build` into `build_module`. Thus one can better reason about what happens plus we can drop some duplication.

scaronni:
> Seems like we're trying to sign the modules twice atm - once in `do_build` and second time in `sign_build`. The latter cannot work in most cases, since it's looking for uncompressed kernel modules - where most distros compress them.
> 
> I'm inclined that we should nuke `sign_build` and inline `prepare_build` `do_build` and `clean_build` into `build_module`. Thus one can better reason about what happens plus we can drop some duplication.

Fine to me, would you do it in a separate merge request? I also created this issue as a reminder for myself, maybe you can attach to it with a separate merge request:

https://github.com/dell/dkms/issues/162

Also, I don't have project access to change members, otherwise I would give you direct commit access. Thanks!

evelikov:
Yup, will do separate MR maybe tomorrow. Mind you I don't know much about module signing, so I won't be tackling #162 anytime soon.

Thanks for the vote of confidence - wrt commit access. I think it's better to use PRs, regardless if one has direct commit access.

PR #175, opened 2021-11-12
Fix creation of temp build dir

heftig:
Since `$source_dir` is just a symlink to the actual source directory, using `cp -a` on it just copies the symlink instead of the contents.

This makes the build fail when the source directory isn't writable.

Commit d038a604bf525b01d7bd12b98706b1f861e4ea27 broke this when it dropped the trailing slash.

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -872,7 +872,7 @@ prepare_build()
 
     # Set up temporary build directory for build
     rm -rf "$build_dir"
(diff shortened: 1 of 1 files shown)


evelikov:
Thanks for the fix @heftig - it's spot on.

Looking through the commit - there are two other instances where the trailing / is dropped - although they are cosmetic print statements.

PR #183, opened 2021-11-19
support CC=clang and LD=ld.lld

phoepsilonix:
1. In the case of kernels built with clang/llvm and lld, clang/llvm and lld are also used to compile kernel modules.
2. Fixing typos ${make_command}

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -571,6 +571,16 @@ read_conf()
         make_command="${MAKE[$index]}"
     done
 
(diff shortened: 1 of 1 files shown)


scaronni:
Why another pull request for the same topic as in https://github.com/dell/dkms/pull/169?

Also, as described in the above pull request, we're trying the approach of checking the `.config` file first.

PR #186, opened 2021-11-20
test: generalize matching lines with several dots

fishilico:
dkms sometimes invokes commands in background and wait for them by display a dot once every 3 seconds (this logic is implemented in function `invoke_command`). This makes it difficult to compare the output of the command with a reference one.

In the tests, replace every sequence of 4 dots or more with 3 dots, when they appear at the end of line. This should fix issues such as the one which occurred in https://github.com/dell/dkms/runs/4249192349


scaronni:
Many thanks!

PR #188, opened 2021-12-10
Revert 122 obsoleteby

InterceptorF:
testing an atto card


scaronni:
I assume this is a mistake, there are tons of commits from years ago in this pull request.

PR #241, opened 2022-09-08
Use inclusive language "correctness check"

julian-klode:
I just ran sed s#sanity#correctness#g on all files.

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -648,9 +648,9 @@ get_module_verinfo(){
     done < <(modinfo $1)
 }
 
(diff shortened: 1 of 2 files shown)


evelikov:
Pardon my ignorance - what's the issue with sanity, do you have some reference material for us to read on?

Bear in mind that sanity and correctness are not at synonymous.

evelikov:
Renaming "sanity" is fine, as long at the result is logically and grammatically correct. Running a random "sed" command, isn't particularly useful.

Feel free to reopen if you're willing to update this.

PR #246, opened 2022-09-10
dkms: Change deprecated egrep for grep -E

Foxboron:
New versions of grep is going to issue a warning when `egrep` is used.

    egrep: warning: egrep is obsolescent; using grep -E

This changes the invocation to `grep -E`

Signed-off-by: Morten Linderud <morten@linderud.pw>


evelikov:
Indeed, introduced with grep 3.8.

Looks like the CI is failing due to unrelated reasons, so let's merge this trivial change. Will have a look at the CI tomorrow

PR #247, opened 2022-09-16
Makefile: make gzip man page reproducible

jelly:
By default gzip records the timestamp so when reproducing dkms this is a problem.

Signed-off-by: Jelle van der Waa <jelle@archlinux.org>

The non-reproducible output can be observed here https://reproducible.archlinux.org/api/v0/builds/338097/diffoscope


evelikov:
Been meaning to redo it, thanks for beating me to it.

Fwiw I've done this in the past, alas @scaronni reverted it https://github.com/dell/dkms/pull/163#issuecomment-939514521 for reasons, which do not seem relevant IMHO.

PR #257, opened 2022-10-17
Add "Tuxedo" distributor id

Matombo:
TUXEDO OS 1 is based on Ubuntu but modifies /usr/lib/os-release. DKMS however should do the same things as on Ubuntu however. In most places this can be archived by the ID_LIKE field, however in the common postinstall script, not the ID_LIKE field is checked, but the ID field.

This patch adds the Tuxedo ID to the one if-case where it is relevant.

Signed-off-by: Werner Sembach <wse@tuxedocomputers.com>


evelikov:
I'm a Debian/Ubuntu noob, so perhaps a silly question: Why don't we use `os-release` in the `dkms_common.postinst` script but  instead we build on the legacy (I believe) `lsb_release` one? Can we bribe into doing that?

Nit: please prefix the commit with the file that's changed - `dkms_common.postinst: bla bla`

Matombo:
> I'm a Debian/Ubuntu noob, so perhaps a silly question: Why don't we use `os-release` in the `dkms_common.postinst` script but instead we build on the legacy (I believe) `lsb_release` one? Can we bribe into doing that?

the `lsb_release -s -i` is actually parsing /usr/lib/os-release if it exists

/etc/lsb_release also exists on Ubuntu, and is parsed directly in another part in DKMS. There however not only the ID but also the ID_LIKE variable is checked, that's why it is no problem.

> Nit: please prefix the commit with the file that's changed - `dkms_common.postinst: bla bla`

Done

Matombo:
Ubuntu is using this python based implementation of the lsb_release command: https://salsa.debian.org/debian/lsb

by using os-release directly do you mean manually parsing /usr/lib/os-release?

Matombo:
Ok reading the manual https://www.freedesktop.org/software/systemd/man/os-release.html standard is to only check /etc/os-release (which may or may not be a symlink to /usr/lib/os-release)

The python implementation of lsb_release i linked (and ubuntu is using) is violating that as it acesses /usr/lib/os-release directly.

Matombo:
Maybe a better proposal, unifing the loggic in dkms a little bit: https://github.com/dell/dkms/pull/270

evelikov:
Looking at the details: we're overriding the ARCH field, which seems to be a non-issue. Namely it is used effectively as a maker - it can be anything "amd64" "x86_64" or even "potatoe". Ubuntu and Debian share the arch naming convention yet Debian is not handled above.

Code was added to fix https://bugs.launchpad.net/ubuntu/+source/dkms/+bug/497149, but admittedly I don't grasp Debian/Ubuntu enough to judge if this is a genuine issue/fix.

@anbe42 (feel free to tag other Debian devs) in your experience, have you ever seen similar patch for dkms in Debian? Does the issue sound something relevant to Debian - aka should we also add you lot in the "amd64" -> "x86_64" mapping, alternatively does it look like something which should be dropped?

evelikov:
I've proposed removing the common.postinst script with https://github.com/dell/dkms/pull/339. If you think that may cause issue please reply in there.

If we end up removing it, this MR will no longer be needed.

PR #258, opened 2022-10-17
refactor: Use `$()` instead of legacy backticks

hyperupcall:
In the same style as https://github.com/dell/dkms/commit/5d35dd150d315d25ef8f6f15b64ffcd444869db6 and https://github.com/dell/dkms/commit/3aafbbb6485f059c9160cb9f8681b3db92974e18, this removes the legacy backtick syntax in favor of `$()`


evelikov:
Thanks for the fix o/

PR #259, opened 2022-10-18
Correct searching for rpm in PATH

hadogenes:
Signed-off-by: Jacek Szafarkiewicz <szafar@linux.pl>

### DIFF
--- a/dkms_common.postinst
+++ b/dkms_common.postinst
@@ -91,7 +91,7 @@ get_newest_kernel() {
         CURRENT_ABI=${CURRENT_ABI%%-*}
         NEWEST_KERNEL=$(_get_newest_kernel_debian "$CURRENT_VERSION-$CURRENT_ABI")
 
-    elif [ $(which rpm >/dev/null) ]; then
+    elif which rpm &>/dev/null; then
         # If RPM based
         NEWEST_KERNEL=$(_get_newest_kernel_rhel)
     fi


evelikov:
The commit message is very misleading - the searching is the same, the output redirection is changed.
Something like "dkms: redirect stderr to null during rpm detection" would be a better fix. @hadogenes can you please re-spin?

xuzhen:
The ```elif [ `which rpm 2>/dev/null` ]; then``` just checks if rpm file exists. We can use ```elif which rpm &>/dev/nul; then``` or ```elif command -v rpm &>/dev/nul; then``` instead

hadogenes:
> The commit message is very misleading - the searching is the same, the output redirection is changed. Something like "dkms: redirect stderr to null during rpm detection" would be a better name. @hadogenes can you please re-spin?

But ```[ `which rpm >/dev/null` ]``` is always false (`[ ]` test is checking if string is not empty and when we add `> /dev/null`, then test always false)

evelikov:
@hadogenes sorry for misleading you there - you're right this fixes the check, but as a whole the existing code is subpar. Can you please follow [xuzhen's advice](https://github.com/dell/dkms/pull/259#issuecomment-1283290237) - thanks in advance.

evelikov:
Fixed up myself, thanks for the great suggestion xuzhen.

PR #261, opened 2022-10-19
tweak: Remove unnecessary glob match test

hyperupcall:
The case is already covered by `-*`


evelikov:
Thanks - looking good. Merging

PR #263, opened 2022-10-24
Fix using PKCS#11 engine as private key

xuzhen:
1. sign-file supports ```pkcs11:*```, not ```*:*```
2. openssl don't support using ```pkcs11:*``` for the -keyout argument

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -916,11 +916,12 @@ prepare_signing()
     fi
     echo "Public certificate (MOK): $mok_certificate"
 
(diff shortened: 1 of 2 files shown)


evelikov:
Should we also update the man page aka [dkms.8.in](https://github.com/dell/dkms/blob/master/dkms.8.in)?

xuzhen:
> Should we also update the man page aka [dkms.8.in](https://github.com/dell/dkms/blob/master/dkms.8.in)?

Of course we should. The framework.conf part in the manpage lacks description of many options, and so dose the output of ```dkms --help```. I will fix those together later.

PR #270, opened 2022-10-26
DRAFT: Avoid using the potentially non standard compliant lsb_release command

Matombo:
Alternative to https://github.com/dell/dkms/pull/257

Check the distribution the way dkms.in does it and not solely rely und the lsb_release command.

### DIFF
--- a/dkms_common.postinst
+++ b/dkms_common.postinst
@@ -105,6 +105,53 @@ get_newest_kernel() {
     echo $NEWEST_KERNEL
 }
 
+# Grab distro information from os-release.
+# Falls back to LSB for compliant distros.
+# Copied from dkms.in
+distro_version()
+{
+    if [[ -r /etc/os-release ]]; then
+        . /etc/os-release
+        if [[ "$ID" = "ubuntu" ]]; then
+            # ID_LIKE=debian in ubuntu
+            echo $ID
+        elif [[ ${#ID_LIKE[@]} != 0 ]]; then
+            echo ${ID_LIKE[0]}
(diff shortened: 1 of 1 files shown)


evelikov:
(Copy/paste for cross-referencing purposes)

I've proposed removing the common.postinst script with https://github.com/dell/dkms/pull/339. If you think that may cause issue please reply in there.

If we end up removing it, this MR will no longer be needed.

bac0n:
I have redone my `distro_version` function (original only check from the beginning of the `running_distibution=` string):

```bash

distro_version(){
    local ID ID_LIKE
    for f in /etc/os-release /usr/lib/os-release; do
        [[ -r $f ]] && ( . "$f" && echo "${ID_LIKE:-$ID}" ) && return
    done
    die 4 "System is missing os-release file."
}

running_distribution=$(distro_version) || exit $?
```
and the case switches:
```bash
case " $running_distribution " in
    *\ @(ubuntu|debian)\ *)
        echo "Matching debian" ;;
    *\ pop\ *)
        echo "Matching pop!_os" ;;
    *)
        echo "no match!" ;;
esac
```
If you want to just match from the beginning of `ID`, `ID_LIKE` as it is now, probably simpler to  cast `ID_LIKE` to an array first:
```bash
ID_LIKE=($ID_LIKE); echo "${ID_LIKE[0]:-$ID}"
```
 that would remove the need for `word*` wildcard matching:
```bash
case "$running_distribution" in
    ubuntu|debian) echo match ;;
esac

PR #272, opened 2022-10-27
Fix --config option

xuzhen:
The --config option didn't participate in the CC & LD checking. And #269 also forget it.


evelikov:
Very nice find.

Small request - in the future please include the "The --config option didn't participate in the..." text within the git commit itself. Thanks o/

Issue #31, opened 2017-09-09
install/uninstall inconsistent


jasmin-j:
Hi!

While testing my new media-build-dkms (https://github.com/jasmin-j/media_build/tree/dkms-master), I discovered several issues concerning the install/uninstall commands.

First of all dkms uses the version got from command modinfo to decide, if a newly built module is newer or equal than the existing one. I am not sure if this is really sufficient.
In my case there are several drivers which didn't increase the version, but the underlying media-core has been changed heavily. So they haven't been installed, although they should. When I want to solve this, I would need "--force". But the "dkms_common.postinst" has no possibility to add "--force" by e.g. an environment variable.
Even more, "template-dkms-mkdeb/debian/postinst", prefer the system wide installed "common.postinst" script. So there is no possibility for a package maintainer to add the "--force" option, other than writing his own postinst script.

I would suggest to add a possibility to add options to the dkms commands used in "common.postinst" preferably by environment variables, because they can be easily set in the package "debian/postinst" file or even at the command line of the "apt-get ..." command.

When it comes to package removing, dkms checks if the installed module and the one from the dkms module directory (/var/lib/dkms/<module>/<version>/<kernel>/<arch>/module) are the same. In case when the module was not installed, "find_actual_dest_module_location" will return an empty string which will let variable "real_dest_module_location" empty, which will then lead to an error of the "rm -f ..." command:

    cx24117.ko:
     - Uninstallation
       - Deleting from: /lib/modules/3.13.0-117-generic/
    rmdir: failed to remove ‘’: No such file or directory
     - Original module
       - No original module was found for this module on this kernel.
       - Use the dkms install command to reinstall any previous module version.

I see here some problems.
a) It is not a good idea to check the module version when installing, but the whole file when uninstalling. This is inconsistent. I don't know if it would be possible to diff the modules at the install step also, instead of checking the version. This would also solve the problem mentioned in the beginning. On the other hand, the modules might be always different, if there is a date or time stored in the module .ko file, which I don't know.

Another possibility would be to check in the uninstall function also the versions from modinfo, but this would not uninstall the module, if it was installed with "--force". So this is no solution.

b) When "real_dest_module_location" is empty, no module was installed, so it would be the best to simply do nothing (doing a continue).

BR,
Jasmin

jasmin-j:
In the beginning of the previous message, I wrote about "--force".
[...]

PR #275, opened 2022-10-28
Fix the search for the path of installed modules

xuzhen:
This patch try to fix a bug: if an installed module is not the same as the compiled one, it cannot be really uninstalled.

It's easy to reproduce:
```
sudo dkms add test/dkms_test-1.0
sudo dkms install dkms_test/1.0
echo aaa | sudo tee -a /var/lib/dkms/dkms_test/1.0/$(uname -r)/$(uname -m)/module/dkms_test.ko
sudo dkms uninstall dkms_test/1.0
sudo dkms install dkms_test/1.0
```
The `dkms uninstall` will try to delete `dkms_test.ko` from `/lib/module/$(uname -r)/`, then the last `dkms install` will say the module is already found in kernel and abort the installation. 

This is because `find_actual_dest_module_location` in `do_uninstall` 
https://github.com/dell/dkms/blob/a26751f07c4fae3ad3175995bd297bec05a1002e/dkms.in#L1452-L1456

will compare the installed module to the compiled one. If they are not identical, `find_actual_dest_module_location` won't display the path.
https://github.com/dell/dkms/blob/a26751f07c4fae3ad3175995bd297bec05a1002e/dkms.in#L1404-L1422

Thus, `do_uninstall` will delete a wrong path.

`find_actual_dest_module_location` also appeared in `module_status_built_extra`, which also do a comparison between the installed module and the compiled one. So this patch won't affect `module_status_built_extra`
https://github.com/dell/dkms/blob/a26751f07c4fae3ad3175995bd297bec05a1002e/dkms.in#L1664-L1679

Another place `find_actual_dest_module_location` appeared is in `list_each_installed_module`, https://github.com/dell/dkms/blob/a26751f07c4fae3ad3175995bd297bec05a1002e/dkms.in#L1292-L1306

which is called by `do_install` and `do_uninstall` for weak modules
https://github.com/dell/dkms/blob/a26751f07c4fae3ad3175995bd297bec05a1002e/dkms.in#L1262-L1268
https://github.com/dell/dkms/blob/a26751f07c4fae3ad3175995bd297bec05a1002e/dkms.in#L1443-L1449
I'm not familiar with weak modules, but I don't think this patch will have side effects on them


evelikov:
I believe you're spot on here, but will have another closer look tomorrow.

Out of curiosity: how did you find this issue?

xuzhen:
> Out of curiosity: how did you find this issue?

This is the bug I mentioned in https://github.com/dell/dkms/pull/274#discussion_r1008187409?

The original test case workflow in #274 is: build module and sign using sha512 -> install module -> check modinfo -> rebuild module and sign using sha256 -> check signature -> uninstall module. Then this bug popped up.

xuzhen:
After reading #31, I realized that we should check the module version in `find_actual_dest_module_location`.

scaronni:
@xuzhen @evelikov can we have a look at the open pull requests and merge them? I would like to do another point release. Thanks!

evelikov:
Yup, it's on my todo list for today

PR #278, opened 2022-11-05
Ensure the kconfig file location is set during building

### LINKED ISSUE #266: sha512 signed modules not works
Modules signed with the dkms-3.0.8 are rejected and I end up at the prompt of initramfs. In my case zfs-modul. The modules are signed but not accepted. Neither with my own keys nor the ones created by dkms. I am using Debian SID

The reason for this can be found in the following line.

eval '"$sign_file" sha512 "$mok_signing_key" "$mok_certificate" "$built_module"'

Here I get hash_algo sha512 signed which doesn't work on my box and my
laptop. If I change this to sha256 secure-boot works as desired. Why
this is so I can not say, this is beyond my knowledge. Also if this is a
bug or not I can't say.

Sorry for my crap english :-)

...

xuzhen:
#266 is not fully fixed by #274. This is because the variable `kernel_config` is not set before calling `prepare_signing` if the module was not built before.

Fix #266.

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -970,8 +970,6 @@ prepare_build()
     # If the module has not been added, try to add it.
     is_module_added "$module" "$module_version" || add_module
 
-    set_kernel_source_dir_and_kconfig "$kernelver"
-
     local -r base_dir="$dkms_tree/$module/$module_version/$kernelver/$arch"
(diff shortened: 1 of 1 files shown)


evelikov:
You are a star, thank you :tada:

PR #283, opened 2022-11-18
improve verbosity

LocutusOfBorg:
log_action_msg is a debianism

### DIFF
--- a/dkms_autoinstaller
+++ b/dkms_autoinstaller
@@ -26,10 +26,17 @@ fi
 # We only have these functions on debian/ubuntu
 # so on other distros just stub them out
 if [ ! -f /etc/debian_version ]; then
-    alias log_daemon_msg=/bin/echo
     log_end_msg() { if [ "$1" = "0" ]; then echo " Done. "; else echo " Failed. "; fi; }
-    alias log_action_begin_msg=log_daemon_msg
-    alias log_action_end_msg=log_end_msg
+
+    # inspired from lsb init-functions script
+    log_action_msg_pre () { :; }
+    log_action_msg_post () { :; }
+    log_action_msg () {
+        log_action_msg_pre "$@"
+        echo "$@." || true
+        log_action_msg_post "$@"
+    }
+
 fi
 
 exec="/usr/sbin/dkms"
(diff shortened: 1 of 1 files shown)


scaronni:
Shouldn't this require some more changes here? https://github.com/dell/dkms/pull/283/files#diff-7dcd9a612f636095321f76fd021321cd3019f4130ae3528d5fcee0f74d2c0a9bL26-L33

LocutusOfBorg:
which changes sorry?

xuzhen:
> which changes sorry?

https://github.com/dell/dkms/blob/ce3e217f7d3746ab23cedf86613f636780b23a72/dkms_autoinstaller#L28-L33

evelikov:
Precisely ^^ - the `log...` is an alias for `echo` for non-Debian. If there is a better alternative do send a PR, but properly sanity check across distros.

LocutusOfBorg:
@evelikov @xuzhen @scaronni thank you! I admit I didn't know where that function came from in Debian.
I implemented a stub of the function for non-Debian systems, so we can use the same function everywhere.
What do you think? this approach looks cleaner to me

evelikov:
Can you share and example of the changes this brings - a before/after picture would be nice. Also please squash the commits.

LocutusOfBorg:
```
     log_end_msg() { if [ "$1" = "0" ]; then echo " Done. "; else echo " Failed. "; fi; }
    # inspired from lsb init-functions script
    log_action_msg_pre () { :; }
    log_action_msg_post () { :; }
    log_action_msg () {
        log_action_msg_pre "$@"
        echo "$@." || true
        log_action_msg_post "$@"
    }


   touch /proc/foo
   log_action_msg "this is a failure"
   echo command that works
   log_action_msg "this one is a success"
``` 

prints a 
```
touch: cannot touch '/proc/foo': No such file or directory
this is a failure.
command that works
this one is a success.
```

LocutusOfBorg:
old version

```
touch: cannot touch '/proc/foo': No such file or directory
this is a failure
command that works
this one is a success
```
prints:
```
touch: cannot touch '/proc/foo': No such file or directory
this is a failure
command that works
this one is a success
```

LocutusOfBorg:
For Debian, no changes in prints since we have the same way we were defined before, for non-debian systems the prints are now looking similar and the code is now shared

LocutusOfBorg:
Hello, ping?

evelikov:
Sorry for the delay. This seems to add an extra full-stop, which does not seem beneficial - so personally I'll abstain.

The other two maintainers (commented above) are welcome to merge it if they choose to.

evelikov:
It seems that the issue wasn't about verbosity or Debian vs others. But a case of missing `\n`.

`log_daemon_msg` should not include the newline, since it's paired with `log_end_msg` which provides it. Relatedly `log_action_msg` should include the newline, since it'a standalone message.

Hats off to anbe42 for the background and fix with  https://github.com/dell/dkms/pull/315

Closing this PR, since it's superseded by the above.

PR #291, opened 2022-12-29
Fix warnings for modules with no version

### LINKED ISSUE #290: module version check causes WARNING build vs. install
The changed module version code in v3.0.9 returns mismatched versions on the same file.  (Alma 8.6, but there are reports from other  distros with the same problem).

The source of the problem:
get_module_verinfo() 
   echo ""  (since there is no 'version' line in modinfo)
   echo "A8869EFB63B2CF973E19263"   (srcversion hex string)

compare_module_version reads these two lines into an array for each file, but the test:
if [[ ! "$ver1" ]] || [[ ! "$ver2" ]]; then
        echo "?"
        return 1
    fi

echos "?" with return status of "no match" , because the  [[ ! "$ver1" ]]  test results in 'true' for the blank first entry in the array.

Suggested fix to test: 
if [[  "${#ver1[\*]}" == 0 ]] || [[  "${#ver2[\*]}" == 0 ]]; then

that checks that the arrays have some entries.

xuzhen:
If there is no `MODULE_VERSION()` macro used in a module's source code, `modinfo` will not show the `version: xxxx` field for the compiled binary. 

if `MODULE_VERSION("")` is used, the situation is similar. `modinfo` will only show `version:` for the compiled binary. 

In both cases, `dkms status` will report `WARNING! Diff between built and installed module!` and `dkms uninstall` will say `Module was not found within /lib/modules/$kernelver/` and the the module will not be removed.

Fix #290
`

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -690,10 +690,6 @@ compare_module_version()
 {
     readarray -t ver1 <<< "$(get_module_verinfo "$1")"
     readarray -t ver2 <<< "$(get_module_verinfo "$2")"
-    if [[ ! "$ver1" ]] || [[ ! "$ver2" ]]; then
-        echo "?"
-        return 1
-    fi
     if [[ "${ver1[0]}" = "${ver2[0]}" ]]; then
         if [[ "${ver1[1]}" = "${ver2[1]}" ]]; then
             echo "=="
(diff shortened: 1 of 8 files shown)


freswa:
I can confirm this fixes #290 for Arch Linux too.

scaronni:
Fixes the same issue on Fedora too.

PR #294, opened 2023-01-22
Fix building with clang and lld

hadogenes:
when building zfs dkms module it fails because KERNEL_CC and KERNEL_LD wasn't set correctly

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -1036,8 +1036,8 @@ prepare_build()
             $"Check $build_dir for more information."
     done
 
-    if [[ -e $kernel_source_dir/.config ]]; then
-        cc=$(sed -n 's|^CONFIG_CC_VERSION_TEXT="\([^ ]*\) .*"|\1|p' $kernel_source_dir/.config)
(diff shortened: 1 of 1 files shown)


xuzhen:
And please change the two variables `cc` and `ld` into local variables.

hadogenes:
@xuzhen corrected

evelikov:
@hadogenes can we interest you in adding zfs into our testing matrix? @xuzhen is doing some amazing work fixing all the CC/LD detection and it would be great if we don't break zfs-dkms.

hadogenes:
It would be great idea, but I don't know how to get kernel compiled with clang.
Maybe ArchLinux and AUR, but it would take a lot of time to compile kernel and than dkms CI.

evelikov:
@hadogenes you can use the [CI from this MR](https://github.com/dell/dkms/pull/298) as reference:
 - existing tiny/clang kernel build/install - perhaps a trimmed down version?
 - get the the zfs code - git clone/curl/wget tarballs?
 - dkms add/build/install
 - dkms status + some sanity checks as you see fit

PR #310, opened 2023-02-26
run_tests: saner handling of sig key/hashalgo

evelikov:
Namely request the field we're interested in.

Aka use `modinfo -F foobar` instead of `modinfo | grep foobar| sed 's/foobar://'` and alike.

@xuzhen can you double-check I'm not missing anything?

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -279,7 +279,7 @@ dkms_test/1.0: added
 EOF
 if ! [[ -d /usr/src/dkms_test-1.0 ]] ; then
     echo >&2 'Error: directory /usr/src/dkms_test-1.0 was not created'
-    return 1
+    exit 1
 fi
 
 echo 'Adding the test module again (expected error)'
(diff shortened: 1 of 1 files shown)


evelikov:
Hmm stumbled upon the kmod built without openssl. Seems like the proper fix is to actually check  for the algo first - if unknown, then skip the key check. The current tests (at least) seem to be backwards.

evelikov:
Pardon for the noise - trying to track down which piece of the stack is borked and thus srcversion is not produced. The original comment was saying Debian, although I suspect it could be kernel version related ... or patch thereof.

kmod itself doesn't even know about that field.

evelikov:
The srcversion seems semi-magical and semi-broken atm - opened an issue for it https://github.com/dell/dkms/issues/311.

The updated PR, should do a saner handing of sig key/hashalgo now ... even though it started with a simple modinfo -F simplification :sweat_smile:

xuzhen:
I think it's ready to merge.

Issue #305, opened 2023-02-15
Signature problem using Fedora Kernel Copr 5.15 Lts


enigma131:
Hi,
sudo modprobe -v lnbp21
insmod /lib/modules/5.15.92-200.fc37.x86_64/extra/lnbp21.ko.xz 
modprobe: ERROR: could not insert 'lnbp21': Key was rejected by service

Key is enrolled  correctly on mainboard and is working on 2 others partitions on same computer (Redhat8 and Debian 11)
Without secure mode modules are loading and working without problem.

repported also at end : https://bugzilla.redhat.com/show_bug.cgi?id=2139560
I use kernel long term 5.15 from https://copr.fedorainfracloud.org/coprs/kwizart/kernel-longterm-5.15/packages/

enigma131:
I've upgraded kernel. 
Still same problem, dkms build the modules with signature, but at reboot is rejected by Dkms service under secure boot.
A way to solve the problem ?

xuzhen:
Before the `machine` keyring was introduced with kernel 5.18, all MOK keys were stored in the `platform` keyring. You can see them with `sudo keyctl list %:.platform`. The only problem is, the `platform` keyring is not used for module signature verification.

Fedora and many other distributions used an out of tree patch to allow the keys contained within the `platform` 
keyring to be used for module verification.

Here is the relevant part of file patch-5.15-redhat.patch within https://kojipkgs.fedoraproject.org//packages/kernel/5.15.18/200.fc35/src/kernel-5.15.18-200.fc35.src.rpm
```
diff --git a/kernel/module_signing.c b/kernel/module_signing.c
index 8723ae70ea1f..fb2d773498c2 100644
--- a/kernel/module_signing.c
+++ b/kernel/module_signing.c
@@ -38,8 +38,15 @@ int mod_verify_sig(const void *mod, struct load_info *info)
 	modlen -= sig_len + sizeof(ms);
 	info->len = modlen;

-	return verify_pkcs7_signature(mod, modlen, mod + modlen, sig_len,
+	ret = verify_pkcs7_signature(mod, modlen, mod + modlen, sig_len,
 				      VERIFY_USE_SECONDARY_KEYRING,
 				      VERIFYING_MODULE_SIGNATURE,
 				      NULL, NULL);
+	if (ret == -ENOKEY && IS_ENABLED(CONFIG_INTEGRITY_PLATFORM_KEYRING)) {
[...]

PR #312, opened 2023-03-11
dkms: skip module signing when hash is unknown

evelikov:
Currently we'll attempt to sign the modules, even when we don't have the
kernel config and/or when CONFIG_MODULE_SIG_HASH is not set.

This serves little purpose, since in either of those cases the kernel
won't check the signature.

As [mentioned previously](https://github.com/dell/dkms/issues/305#issuecomment-1445943045)

@xuzhen please give this a once-over. I suspect that some tests may fail, because we duck-tap (read workarounds) which need updateing


evelikov:
As anticipated, there were some holes in the behaviour and tooling. Aka we were trying to sign things on Debian 9, which does not support it.

While there, I've also tweaked the tests to strip/generalise _only_ as needed.

PR #315, opened 2023-03-22
Misc cleanup

anbe42:
as few bits taken out of the autoinstall-fixes branch


run_test.sh
```diff
@@ -14,6 +14,9 @@ echo "Using kernel ${KERNEL_VER}/${KERNEL_ARCH}"
 PATH="$(pwd):$PATH"
 export PATH
 
+# Avoid output variations due to parallelism
+export parallel_jobs=1
+
```
evelikov, line 15: Please apply `s/parallel_jobs=1/MAKEFLAGS="-j1"/;s/make -j1/make/` and let's merge these. Thanks

anbe42: Using MAKEFLAGS doesn't work due to the way how dkms injects the -j flag:
```
local the_make_command="${make_command/#make/make -j$parallel_jobs KERNELRELEASE=$kernelver}"
```

evelikov: Could swear that we were honouring the parallelism in `MAKEFLAGS` instead of blindly using `parallel_jobs`.
Looking at it again - we don't. It's a bug we should probably fix (will try and tackle this over the weekend).

Thanks for the patience

PR #318, opened 2023-03-25
CI: Update distros

evelikov:
Going through the distributions list:
 - remove: Alpine 3.14 and earlier (3.14 is EOL next month)
 - add: Apline 3.15 - 3.17
 - keep: CentOS 7 is EOL no newer version officially has docker images (could be wrong here)
 - keep: Ubuntu 18.04 (EOL this month) - should we drop this @xnox @anbe42
 - keep: Debian 9 (EOL 2020, LTS 2022) - planning to keep until Debian 12 is released in a few months
 
 @xuzhen @scaronni should we add some Suse/Fedora/other versions in the mix?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,8 +23,8 @@ jobs:
           - {name: "archlinux", tag: "latest", variant: "-lts"}
           - {name: "archlinux", tag: "latest", variant: "-zen"}
           - {name: "centos", tag: "7"}
-          - {name: "almalinux", tag: "8"}
           - {name: "almalinux", tag: "9"}
+          - {name: "almalinux", tag: "8"}
           - {name: "debian", tag: "11"}
           - {name: "debian", tag: "10"}
           - {name: "debian", tag: "9"}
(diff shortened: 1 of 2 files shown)


evelikov:
Looking at the failures, a few interesting bits became apparent:
 - the good: tests have a hard-coded distro/module compression list - some distros are not tested in CI (fedora/rhel/ovm/suse/sles/opensuse)
 - the good: Alpine has switched to .gz compressed modules \o/
 - the bad: vast majority of distros do not compress the kernel modules
 - the ugly: module compression detection is broken - relies on the potentially missing `modules.dep` (instead of the CONFIG)
 - the ugly: we mostly get away with ^^ since we always check for both compressed and uncompressed modules

evelikov:
Let's go with a quick fix for now by adding the `.gz` expectation for Alpine.

I'm moderately contemplating about ripping out some/all of the custom installation (which includes the module stripping/signing) in favour of `make module_install`. On a quick testing it just works and does the right thing.

@scaronni do you recall why `module_install` isn't used in the first place? Do you have any particularly complex/convoluted module that confuses dkms?

Issue #244, opened 2022-09-09
Module signing is broken in latest dkms on debian


blastrock:
Hi,

I've been using dkms signing capabilities with sign_tool and the sign_helper script, but the last update of my debian unstable broke it all.

```
Sign command: /usr/lib/linux-kbuild-5.19/scripts/sign-file
Signing key: /root/dkms.key
Public certificate (MOK): /root/dkms.der
Binary update-secureboot-policy not found, modules won't be signed
```

There seem to be multiple issues with the script, or maybe I'm not using it like I'm supposed to, please tell me.

1. There is a typo here https://github.com/dell/dkms/blob/49d647b6476d605feb10548e33fc3807415ec724/dkms.in#L893 . We check for kmodsign and then print kmod-sign in the error message.
2. These two conditions look reversed to me https://github.com/dell/dkms/blob/49d647b6476d605feb10548e33fc3807415ec724/dkms.in#L892-L900 . They should be `if [[ ! -x ...`
3. We expect debian to have kmodsign https://github.com/dell/dkms/blob/49d647b6476d605feb10548e33fc3807415ec724/dkms.in#L890 , but I definitely can't find it on my debian, isn't that a ubuntu thing? There is this issue opened https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=939392 about that.
4. And finally, I don't get why we create a new key and enroll it on every run here https://github.com/dell/dkms/blob/49d647b6476d605feb10548e33fc3807415ec724/dkms.in#L905-L906 . The keys I sign with are already enrolled, why would that be necessary?

xuzhen:
2\. This bug was fixed by https://github.com/dell/dkms/pull/242

4\. update-secureboot-policy won't create new key if /var/lib/shim-signed/mok/MOK.der exists, and won't reenroll it.
```
'--new-key')
    create_mok
    exit 0
    ;;

'--enroll-key')
    if [ -e "$SB_KEY" ]; then
        if mokutil --test-key "$SB_KEY" | \
                grep -qc 'is not'; then
            do_enroll=1
        fi
    else
        echo "No MOK found."
        exit 1
    fi
    ;;
```
```
create_mok()
{
    if [ -e "$SB_KEY" ]; then
        return
    fi

    echo "Generating a new Secure Boot signing key:"
    openssl req -config /usr/lib/shim/mok/openssl.cnf \
        -subj "/CN=`hostname -s | cut -b1-31` Secure Boot Module Signature key" \
[...]

PR #322, opened 2023-03-26
dkms: reinstate --enroll-key

evelikov:
This is a Debian/Ubuntu patch, the following is their original commit message:

 When a brand new secureboot key is created, and it hasn't been
 previously enrolled as a mok key, it will be rejected by the
 kernel. After creating a new key, one should be enrolling it.

Side note: The update-secureboot-policy script seemingly originates from Ubuntu, where Debian has an ancient copy, which lacks both --new-key and --enroll-key.

Fixes: 3ca52f8 ("Fix signing on ubuntu & debian (#244)")

@xuzhen @anbe42 jfyi

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -166,6 +166,8 @@ genericize_expected_output() {
     fi
     # Signing related output. Drop it from the output, to be more generic
     if (( NO_SIGNING_TOOL == 0 )); then
+        sed -i '/^EFI variables are not supported on this system/d' ${output_log}
+        sed -i '/^\/sys\/firmware\/efi\/efivars not found, aborting./d' ${output_log}
         sed -i '/^Sign command:/d' ${output_log}
         sed -i '/^Signing key:/d' ${output_log}
         sed -i '/^Public certificate (MOK):/d' ${output_log}
--- a/run_test.sh
+++ b/run_test.sh
@@ -10,6 +10,10 @@ KERNEL_VER="${KERNEL_VER:-$(uname -r)}"
 KERNEL_ARCH="$(uname -m)"
 echo "Using kernel ${KERNEL_VER}/${KERNEL_ARCH}"
 
+# debconf can trigger at random points, in the testing process. Where a bunch of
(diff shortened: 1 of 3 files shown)


xuzhen:
If we're going to do this on Ubuntu, what about other distros? The behavior of dkms on each distribution should be as consistent as possible.

And before https://github.com/dell/dkms/commit/3ca52f8769bdf7ebdc83735355fff7c5c0664152, dkms will checked if secureboot was enabled before enrolling the key. Without those code, `update-secureboot-policy --enroll-key` will show unnecessary errors in output on platforms that do not enable or support secureboot.

run_test.sh
```diff
@@ -10,6 +10,10 @@ KERNEL_VER="${KERNEL_VER:-$(uname -r)}"
 KERNEL_ARCH="$(uname -m)"
 echo "Using kernel ${KERNEL_VER}/${KERNEL_ARCH}"
 
+# debconf can trigger at random points, in the testing process. Where a bunch of
+# the frontends cannot work in our CI. Just opt for thee noninteractive one.
```
xuzhen, line 14: thee -> the

evelikov:
Welcome back @xuzhen p/

Yes, ideally dkms will have _zero_ knowledge about distribution specifics.

From a quick look the vast majority are caused by our open-coding of the installation/copy/strip/sign stages. Off the top of my head all of that can be avoided by using the kernel `module_install` rule. I am itching to give that a try, although it has a much greater chance of causing issues - hence the quick fix here.

The patch is taken as-is from the Debian repo, so it should not cause (too many) errors. CI also looks happy so we can fix any issues as follow-up.

xuzhen (changes requested):

xuzhen:
I think we should mention this special handling for Ubuntu in README.md

run_test.sh
evelikov: Thanks fixed.

evelikov:
Looking at the README - it had a few typos + was misleadingly mentioning Debian. Thanks o/

PR #335, opened 2023-06-19
Makefile: fix extension of kernel-install plugin

AndrewAmmerlaan:
According to the kernel-install manual plugins need to have the `.install` extension, otherwise they are ignored:

```
kernel-install will run the executable files ("plugins") located in the directory
/usr/lib/kernel/install.d/ and the local administration directory /etc/kernel/install.d/. All files
are collectively sorted and executed in lexical order, regardless of the directory in which they
live. However, files with identical filenames replace each other. Files in /etc/kernel/install.d/
take precedence over files with the same name in /usr/lib/kernel/install.d/. This can be used to
override a system-supplied executables with a local file if needed; a symbolic link in
/etc/kernel/install.d/ with the same name as an executable in /usr/lib/kernel/install.d/, pointing
to /dev/null, disables the executable entirely. Executables must have the extension ".install";
other extensions are ignored.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ install: dkms dkms.8
 	mkdir -p $(ETC)/framework.conf.d
 	install -D -m 0644 dkms.bash-completion $(BASHDIR)/dkms
 	install -D -m 0644 dkms.8 $(MAN)/dkms.8
-	install -D -m 0755 kernel_install.d_dkms $(KCONF)/install.d/dkms
+	install -D -m 0755 kernel_install.d_dkms $(KCONF)/install.d/40-dkms.install
 	install -D -m 0755 kernel_postinst.d_dkms $(KCONF)/postinst.d/dkms
 	install -D -m 0755 kernel_prerm.d_dkms $(KCONF)/prerm.d/dkms
 	gzip -n -9 $(MAN)/dkms.8


evelikov:
Thanks for the PR. How did you notice the issue?

While I agree we should follow the documentation, I'm weary about regressions. Can you check and provide some references that this won't cause (m)any issues and/or that distros are already renaming the file.

For example: Arch [does not install](https://gitlab.archlinux.org/archlinux/packaging/packages/dkms/-/blob/main/PKGBUILD#L68) the files, as it can be seen from the [package contents](https://archlinux.org/packages/extra/any/dkms/files/)

AndrewAmmerlaan:
> Thanks for the PR. How did you notice the issue?

Installing a new kernel did not automatically trigger the rebuilding of dkms modules. Of course after rebooting the systemd service still takes care of this, so not a huge problem. 

- Fedora (and by extension anything based on it) at least are already working around this, they have a `40-dkms.install` in `/usr/lib/kernel/install.d`. You can download the rpm from here to see what they are installing: https://koji.fedoraproject.org/koji/buildinfo?buildID=2195311

- Debian/Ubuntu do not appear to rename the file, but they might have some other mechanism in place to trigger dkms when the kernel package gets updated.

> For example: Arch [does not install](https://gitlab.archlinux.org/archlinux/packaging/packages/dkms/-/blob/main/PKGBUILD#L68) the files, as it can be seen from the [package contents](https://archlinux.org/packages/extra/any/dkms/files/)

- Yes, Arch seems to be relying on the pacman hooks to achieve the same functionality (kernel gets installed, dkms gets triggered). Note that making the change in this PR won't break anything because I'm pretty sure Arch does not use the kernel-install command to install the kernel in the first place. 

- Gentoo (which I maintain a dkms package for) does not rename the file (yet). The impact of the issue is low so I decided to not adjust the ebuild with a temporary workaround and instead just fix it upstream and wait for the next release.
 
In fact, AFAIK it is pretty uncommon to use the kernel-install command (which should then call this plugin) except on source based distributions like Gentoo or when installing a kernel manually (`kernel-install` gets called when running the kernel's `make install`). So the impact of the issue here is low, nonetheless it is very easy to do it properly and make this work as it was intended.

Worst case scenario you end up running dkms twice, the second time won't do anything since dkms will find that the modules are already installed and skip.


```
KCONF = $(DESTDIR)/etc/kernel
```

By the way, technically this should really be `KCONF = $(DESTDIR)/usr/lib/kernel`. `/usr` is where packages and default configurations are installed. `/etc` is for configuration and scripts managed by the system administrator. Note that making this change will also not break anything since both directories are checked by `kernel-install`, it is important however to prefix the plugin file with some two digit number to control the order in which the plugins are executed. Currently the dkms plugin is executed last because `/etc/kernel` is run after `/usr/lib/kernel`, so to retain this behaviour a suitable number could be 95. Not that it really matters because dkms is all about modules and not the kernel itself so it will work fine in any order.

```
andrew-gentoo-laptop andrew # ls /usr/lib/kernel/install.d
00-00machineid-directory.install  50-depmod.install  50-dracut.install  51-dracut-rescue.install  90-loaderentry.install  90-uki-copy.install
```

evelikov:
Looking at Fedora it has [an extra install snippet](https://src.fedoraproject.org/rpms/dkms/c/8b1b42aee6a7da19dd51b261691997d1b462464f) although the commit message is bit short.

```
    install -p -m 644 -D kernel_install.d_dkms \            
                %{buildroot}%{_prefix}/lib/kernel/install.d/40-%{name}.install
```

@scaronni you seem to be the author, do you recall what inspired that change? Do you think it's safe for us to step even further and use `40-dkms.install` as per your Fedora patch?

@anbe42 can you have a quick check this won't break any Debian tooling? At a glance I could not spot any.

scaronni:
> Looking at Fedora it has [an extra install snippet](https://src.fedoraproject.org/rpms/dkms/c/8b1b42aee6a7da19dd51b261691997d1b462464f) although the commit message is bit short.
> 
> ```
>     install -p -m 644 -D kernel_install.d_dkms \            
>                 %{buildroot}%{_prefix}/lib/kernel/install.d/40-%{name}.install
> ```
> 
> @scaronni you seem to be the author, do you recall what inspired that change? Do you think it's safe for us to step even further and use `40-dkms.install` as per your Fedora patch?

Sorry can't really remember, it's 3 years and 4 months since that was applied. If I'm not mistaken somone pointed out in a bug that was needed to be installed to make it work.

I inherited an abandoned DKMS package in Fedora/EPEL and as well an almost abandoned repository in Github around the same time, too much has passed since.

anbe42:
I don't think this file (although we ship `/etc/kernel/install.d/dkms`) is used at all in Debian. The `linux-image-*` maintainer scripts use `run-parts` (from package debianutils) to run scripts from `/etc/kernel/{pre,post}{inst,rm}.d/` and there is no corresponding location in /usr.

AndrewAmmerlaan:
Alright I changed to commit to install the plugin as is done in Fedora (`/usr/lib/kernel/install.d/40-dkms.install`). 40 will cause the dkms plugin to run before dracut (which is 50), probably this is a good idea in case we want to include dkms modules into the dracut initrd.

Makefile
```diff
@@ -15,7 +15,7 @@ MAN = $(DESTDIR)/usr/share/man/man8
 INITD = $(DESTDIR)/etc/rc.d/init.d
 LIBDIR = $(DESTDIR)/usr/lib/dkms
 BASHDIR = $(DESTDIR)/usr/share/bash-completion/completions
-KCONF = $(DESTDIR)/etc/kernel
+KCONF = $(DESTDIR)/usr/lib/kernel
```
evelikov, line 18: This will cause all 4 scripts to be installed in /usr/lib/kernel which might cause problems:

 - the `kernel_install.d_dkms` has hardcoded `/etc/kernel` references
   - can you please add an extra commit  to this PR to sed `sed s/@FOO@/$KCONF/` the path?
 - Debian has hardcoded lookup for `/etc` and no `/usr` equivalent and bypasses `kernel-install`
   - feel free to follow-up with Debian people with an issue, path, other
 
IMHO having these in under `/usr/lib/kernel` would be great, we cannot have the above line just yet. So let's drop it.

evelikov (changes requested):

Makefile
AndrewAmmerlaan: > the kernel_install.d_dkms has hardcoded /etc/kernel references 

Oops, my bad, I hadn't noticed that the path was hardcoded.

> can you please add an extra commit to this PR to sed sed s/@FOO@/$KCONF/ the path?

I think we can do even better! I have added a commit that makes the path of the postinst and prerm scripts that we call from dkms.install relative. This way they will always work regardless of what we set KCONF to, without resorting to sed.

> IMHO having these in under /usr/lib/kernel would be great, we cannot have the above line just yet. So let's drop it.

I have dropped this part of the patch. Users and packagers can always override it with `make install KCONF='$(DESTDIR)/usr/lib/kernel'` anyway.

evelikov: The runtime expansion is a no-go IMHO for a couple of reasons:
 - tools running as root avoid use full paths, unless it's unavoidable
 - as-is this will break Fedora because they copy the file to `/usr/` while the rest are in `/etc` - yes they can workaround that, but they should not have to

Please use my original suggestion - the Makefile has some examples.

Thank you

AndrewAmmerlaan: > as-is this will break Fedora because they copy the file to /usr/ while the rest are in /etc - yes they can workaround that, but they should not have to

The same would happen if we use sed to adjust the path.

Also it would be trivial for Fedora to adjust the spec file to accommodate for this change. Simply remove the custom install line and override KCONF instead.

evelikov: > The same would happen if we use sed to adjust the path.

I cannot see how that would happen. Can you elaborate?

> I have not heard of this, and besides /etc/kernel/install.d is also a full path....

Yes `/etc/kernel/install.d` is full path. The proposed change (snippet below) isn't.

```bash
SCRIPT=$(readlink -f "$0")
SCRIPTPATH=$(dirname "$SCRIPT")

"$SCRIPTPATH/../postinst.d/dkms" "$2"
```

AndrewAmmerlaan: > I cannot see how that would happen. Can you elaborate?

If we sed the path and then move the script, the path is still wrong.

In any case, your sed solution is not an option because it completely breaks relocatability of the DESTDIR. If we set DESTDIR to some temporary directory, then this directory would end up in the script. Gentoo at least does this since it is required if the package manager is to keep track of what is actually being installed. I strongly suspect other package management systems also rely on relocatability (i.e. make an install image somewhere, and merge this install image to the file system. As opposed to installing to the file system directly). If we start sed'ing DESTDIR into the installed files something is sure to break somewhere.

I would strongly suggest to use my relative path solution instead. In Gentoo we always aim to use relative paths, precisely because we rely on the relocatability of the install image.

Also I fundamentally disagree that when fixing some bug it is possible or even desirable to ensure backwards compatibility with workarounds for that very same bug. It happens all the time that I implement some downstream workaround before attempting to properly fix something upstream. Once the solution is integrated upstream, then I have to remove my downstream workaround. This is a very normal process, installation scripts (ebuild, rpm spec or otherwise) simply have to be adjusted every now and then.

AndrewAmmerlaan: Another option is to simply integrate the postinst and prerm scripts into the main script. WDYT?

evelikov: > If we sed the path and then move the script, the path is still wrong.

That would be an issue if only DESTDIR is embedded.

> Another option is to simply integrate the postinst and prerm scripts into the main script. WDYT?

This will intentionally break Debian and likely other distros.

Sounds like it's better if we drop the path handling in this PR. It's been on my TODO list anyway.

AndrewAmmerlaan: > > If we sed the path and then move the script, the path is still wrong.
> 
> That would be an issue if only DESTDIR is embedded.

Then I don't understand what you want to sed in. I thought you wanted to sed in KCONF, but this contains DESTDIR?

evelikov: Yes, I've naively forgotten than the dkms variables are borked and contain DESTDIR. Sorry for taking you on this wild ride.

The 40-dkms.install commit is fine.

I am looking at fixing the paths... unless my brain melts before hand.

anbe42:
It's very unusual to have `$(DESTDIR)` embedded in the `$(FOODIR)` variables.
Usually an install targets looks more like this:
```
FOODIR = /usr/foo/bar
install:
        install -d $(DESTDIR)$(FOODIR)
        install some.file $(DESTDIR)$(FOODIR)
```
and can be customized via `make install FOODIR=/usr/local/foobar` or `make install DESTDIR=/tmp/installroot` (or both)

Makefile
AndrewAmmerlaan: Aaaah now I understand.

I've dropped the second commit, leaving only the renaming of the plugin.

We can look into adjusting the install paths as @anbe42 suggested later. If we do it that way then a sed won't break the relocatability.

evelikov (approved):

evelikov:
> It's very unusual to have `$(DESTDIR)` embedded in the `$(FOODIR)` variables.

Indeed, which is why I unintentionally misled Andrew for a moment.

Thank you for the MR and patience Andrew :bow:

AndrewAmmerlaan:
:partying_face: It was a slightly mutually confusing ride, but we got there in the end. Thanks for merging!

Issue #324, opened 2023-03-29
Remove hard-coded paths


evelikov:
[Earlier commit}(https://github.com/dell/dkms/commit/1e24a54acaa2e3a8aa6e8e24b21f47553370c4fe) added the full path into the systemd service. That's the correct thing to do, although it did not consider that the path is can be selected at build time - see `grep SBIN Makefile`.

For this task, one should audit the whole repo for SBIN and other hard-coded paths and `sed ... s|#FOO#|$FOO|` replace them, like the others in the `Makefile`

evelikov:
Essentially look at all the `FOO = $(DESTDIR)/tada` instances, applying distribution backed data if the variable is ever overwritten. If so, convert the code/docs/scripts to use actual user-data. Otherwise see if it makes sense to remove the variable or instead use `:=`.

evelikov:
Exhibit A: https://github.com/dell/dkms/pull/335

Namely: not only are we embedding the incorrect path, but we also conflate merge DESTDIR into the respective directories. At a glance I see 3 options, in order of increasing preference:
 - the bad idea: simply drop the `DESTDIR` merging
   - effectively a silence breakage, `make install` will work w/o errors and files are misplaced
 - reasonable compromise : deprecate the old ones, error if set and introduce new
   - `make install`  will error with `SBIN is deprecated, use sbindir ensuring it does not contain DESTDIR` or alike
 - ideal solution: if `DESTDIR` is present within the `SBIN`/etc paths - warn and strip it
    - don't think I can pull that off, likely to be convoluted and/or fragile if we manage to :frowning_face:


Currently leaning toward the second option. Unless someone knows how to get the last one working, or has other suggestions.

Happy to hear the input of @anbe42, @scaronni and others.

anbe42:
I would drop `$(DESTDIR)` from all variable definitions, add it to the `install` targets instead and deliberately break everyone relying on the old broken behavior. Since nobody filed a bug about it, nobody seems to use it.

evelikov:
Are you suggesting the first option - silent breakage? Debian does not set KCONF and friends, but others like Arch and Fedora do.

AndrewAmmerlaan:
> Are you suggesting the first option - silent breakage? Debian does not set KCONF and friends, but others like Arch and Fedora do.

[...]

PR #339, opened 2023-08-15
Remove embedded DESTDIR in make variables, honour them

evelikov:
This series was inspired by a bunch of hardcoded paths across the project. The more i looked into the code, the bugs stared back at me ...

As a whole this series does a few closely related things:
 - Removes a few seemingly broken and unused scripts
 - No longer compresses the man page
 - Removes some should-not-have-been-a-variable and unused-variable instances
 - Removes the uncommon hard-coded DESTDIR prefix for variables
 - Ensures the remaining variables are honoured across the board

In practical terms this means:
 - find-provides, lsb_release and dkms_common.postinst are gone. @anbe42, @xnox, @scaronni others, I believe Debian Ubuntu and Fedora should not be affected by this removal. Can you give it a quick check on your end and let me know?
 - The gzip manpage compression is left for builders/distros - @AndrewAmmerlaan you can drop the workaround from the Gentoo ebuild now
 - setting VAR, ETC, BASHDIR, MAN, DOCDIR and SHAREDIR is no longer allowed - only Arch was setting one of these BASHDIR - to the already default value
 - setting SYSTEMD, SBIN, LIBDIR and KCONF path does not need explicit DESTDIR prefix
 - hardcoded paths are removed in favour of the SBIN, LIBDIR and KCONF variables - @scaronni @AndrewAmmerlaan

Some notes to the Arch maintainer @seblu (will open an Arch issue in a moment):
 - the 4 year old PKGBUILD.26 can be removed
 - the moddir patching aka `s,/lib/modules,/usr/lib/modules,g` should be dropped - we don't patch kmod either
 - BASHDIR can be dropped, it's the default
 - SBIN, KCONF and LIBDIR are relative to DESTDIR so they should be updated
 - `hook.sh` should probably be reworked to use dkms autoinstall - happy to make autoinstall changes to make that possible

Closes: https://github.com/dell/dkms/issues/324


AndrewAmmerlaan:
> * The gzip manpage compression is left for builders/distros - @AndrewAmmerlaan you can drop the workaround from the Gentoo ebuild now
...
> * hardcoded paths are removed in favour of the SBIN, LIBDIR and KCONF variables - @scaronni @AndrewAmmerlaan

Great! Thanks for the heads-up.

anbe42:
I believe the legacy support in dh_dkms can go away (just need to check that it is not used anymore by any package) and the howto should be dropped.

But installation of a *-dkms package calls /usr/lib/dkms/common.postinst from its postinst script to build the newly installed module for all installed kernel headers, so that script is still needed. And it is expected to exist by all existing *-dkms packages that depend on dkms >= 2.1

I'll look into this in more detail after holidays ...

AllKind:
@evelikov 
As I was made aware here https://github.com/openzfs/zfs/pull/15415#issuecomment-1969423968, the way I implemented the dkms routines for ZFS do not work in chroot environment.
The only way to make it work I see is to reintroduce the the use of common.postinst.
So I come here to ask if you have any other idea?

PR #344, opened 2023-09-24
run_test.sh: add gentoo as valid os id in tests

AndrewAmmerlaan:
In Gentoo I find the dkms modules in `kernel/extra` so this is where I added the `| gentoo` option.


AndrewAmmerlaan:
> Can you please add Gentoo to the CI matrix? As-is we cannot realistically check/validate this.

Done!

Issue #320, opened 2023-03-25
Add more distributions in the test matrix


evelikov:
Currently we have a bunch, although a few more won't go amiss. Namely - Fedora, RHEL, CentOS Stream and Suse (opensuse or SLES).

It's likely we don't need all of these, I'm listing them since they're explicitly mentioned across the codebase.

anbe42:
please drop Debian 9 from the test matrix, it has been recently archived

evelikov:
> please drop Debian 9 from the test matrix, it has been recently archived

Done with https://github.com/dell/dkms/pull/331.

I wonder how is ELTS supposed to work, given that one cannot get the packages even?

PR #350, opened 2023-10-11
CI: add centos stream8/9

evelikov:
This unblocks https://github.com/dell/dkms/pull/349 while also part-fixing https://github.com/dell/dkms/issues/320


scaronni:
CentOS Stream has a shorted life span compared to the RHEL rebuilds/forks by almost two years, so DKMS will be maintained for longer in EPEL 8/9 than the life of CentOS Stream 8/9. But as per the other ticket, we can also drop the distribution from the tests sometime prior, shouldn't be a big deal.

Another option is to use AlmaLinux or RockyLinux in place of CentOS Stream 8/9.

PR #377, opened 2023-12-08
completion: Add value completion for more arguments

hyperupcall:
This improves completion by:

- Properly completing directory names and path names for more flags. Basic completion for `-a` was also implemented (defaulting to current architecture)
- Restructure `-@\(c|-spec|-archive|-config\)` to something that works and is simpler. Now, even `extglob` does not need to be enabled for the new completion to work


hyperupcall:
Sounds good, I will get to it if I can

PR #378, opened 2023-12-11
Report failing command in invoke_command()

AllKind:
Report the command that had a bad exit status.
This gives a better picture of what is happening.

Quite often I saw this:

> Cleaning build area...(bad exit status: 2)

Which makes one wonder, what actually failed.

With this patch:

> Building module:
> Cleaning build area...(bad exit status: 2 of: <make -C /lib/modules/6.6.3-100.fc38.x86_64/build M=/var/lib/dkms/zfs/2.2.1/build clean>)

> Running the post_build script:
> Cleaning build area...(bad exit status: 2 of: <make -C /lib/modules/6.6.3-100.fc38.x86_64/build M=/var/lib/dkms/zfs/2.2.1/build clean>)

PR #380, opened 2023-12-13
Report failing command in invoke_command()

evelikov:
This builds on the ideas in https://github.com/dell/dkms/pull/378, namely:
 - replaces the somewhat misleading/inconsistent "make ..." command during module building
 - passes the redirection file/log as argument to the function
 - prints/reports the failing command


AllKind:
Cool!

PR #387, opened 2024-01-25
CI fixes from anbe42

evelikov:
Basically https://github.com/dell/dkms/pull/385 without the first commit.

Thanks again for the fixes @anbe42 sorry it took so long to review/land.

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -271,6 +271,10 @@ case "${os_id}" in
         ;;
 esac
 
+grep "^CONFIG_MODULE_COMPRESS" "${kernel_config}" || true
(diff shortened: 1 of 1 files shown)


AndrewAmmerlaan:
This is caused by my changes to our installkernel scripts, I'll fix it tomorrow.

AndrewAmmerlaan:
I made the changes that should fix it here: https://github.com/dell/dkms/pull/388

But now its failing somewhere else in Debian/Ubuntu complaining about a missing `os-release`, the gentoo job got cancelled because of this and I have therefore not yet verified that it actually works again now.

PR #388, opened 2024-01-26
workflows/tests.yml: fix Gentoo testing

AndrewAmmerlaan:
Add --autounmask-continue to tell portage to make whatever configuration changes it deems necessary, and then continue.

Add --getbinpkg to fetch prebuilt packages from the binary host to speed things up a bit.


AndrewAmmerlaan:
> Must admit I don't quite follow the `--archive` to ` --preserve=all` change. The former is `-dR --preserve=all`, where `-d` is `--preserve=links` and `-R` is recursive were we copy a single file.
> 
> Please add a small note with rational in the commit message.

To be honest, I don't understand why this fixes the problem. Somehow it fails with `--recursive`, but it does not make sense because it should be a single file. In any case we don't need `--recursive` because it should always be a single file.

AndrewAmmerlaan:
> How does Gentoo ignore the kernel config toggle about the module compression? Does Gentoo's tooling not use `make modules_install` or perhaps it's invoked in particular way?

We configure the kernel with support for xz compressed modules, then patch the build system to skip the actual compression in `make modules_install`. This allows us to distribute a kernel that supports both compressed and uncompressed modules, while leaving the modules in the package tarball uncompressed. This is more efficient compression wise because xz compressing a bunch of uncompressed files results in a smaller tarball compared to xz compressing a bunch of files that are each individually already xz compressed. In the install phase the modules are then xz compressed or not depending on whether the user enabled `USE=modules-compress` (the default is disabled).

PR #389, opened 2024-01-26
Consistently use install_tree, allow overriding /lib/modules

evelikov:
With many distributions moving from /lib/modules to /usr/lib/modules,
allow for the builder to select their option. For backwards
compatibility the default is still /lib/modules.

As result distros for example Arch don't need to patch the project.
Plus this nicely aligns with kmod, which recently allowed for similar
change.

To override set the MODDIR variable during make.

@scaronni @anbe42 @seblu happy to hear if you have any concerns about this.

Fwiw kmod has landed /usr/lib/modules support [with this series](https://lore.kernel.org/all/cover.1689681454.git.msuchanek@suse.de/), although the last release v31 was just prior to that work.


scaronni:
I think it's a very good idea! Shall we merge and create a new release with all the recent changes?

Issue #328, opened 2023-04-04
Move distribution specifics into the config


evelikov:
As mentioned previously https://github.com/dell/dkms/pull/323#discussion_r1150398999, ideally dkms will not know (or care) about the distribution it's run against.

Any specifics like `DEST_MODULE_LOCATION` overrides will be set in `/etc/dkms/framework.conf` and the user can override those with the usual `/etc/dkms/framework.conf.d` mechanism.

For this task, we should:
 - construct a list of distro specific derivation(s), for each
 - check if it's applicable - eg. do we need the collisions/archival for non Debian/Ubuntu platforms
 - add a toggle in `framework.conf`, honour that in dkms
 - add make `install-foo` variant which produces creates/uses the correct config for the given distribution

Nice to have:
 - check with existing distros, if they use all the `install-foo` files, update as necessary
   - eg. Arch uses `make install` and manually removes kconf files and libdir

evelikov:
From https://github.com/dell/dkms/issues/329 we're currently running `rpm` indiscriminately which is causing problems. We should either:
 - remove the offending calls, or
 - use a conf knob to control that

evelikov:
Another set of distro specific variations - module signing. In particular:
 - location of `sign-file` and MOK key/cert
 - deviation across Ubuntu/Debian of the update-secureboot-policy tool

At a glance, if we are to use `modules_install` itself aka https://github.com/dell/dkms/issues/319 then we won't need to care about this.

evelikov:
Moving the original module/collisions to standalone location:
 -  on all distros apart from Debian/Ubuntu we backup the original module (on build) and restore it on uninstall

Might be worth tracking down the reason behind this variance and removing it, in the mid/long run. At least initially preserving the behaviour seems like a good first step.

evelikov:
The above are the distro-specific variations we have so far, although others may emerge at some point.

They don't have to be fixed all together/in a single PR. So if anyone is bored and like to write some bash - patches welcome o/

evelikov:
@HanabishiRecca the original is stored outside of the normal search path. Note that the kernel does not care about any of this - kmod does all the path/priority handling as well as loading.

The ultimate goal here is not to fix everything that's off with dkms, but migrate from a bunch of "running_distribution" hacks to a data/config driven flow.

Issue #319, opened 2023-03-25
Switch to `modules_install` or cleanup the module compression handling


evelikov:

scaronni:
As far as I know, this is just historical stuff and has no particular reason to exist in its current form, it dates back to ~2002?

I think the main reason was just supporting old EOL distributions and other OSes (Hurd).

scaronni:
@anbe42 you were involved with:https://github.com/dell/dkms/issues/319

scaronni:
@anbe42 I've give you contributor access to DKMS.

PR #402, opened 2024-02-10
get module_compressed_suffix from .config

masonlherndon:
The old (current) method of parsing `modules.dep` fails to detect compression if none of the currently installed modules are compressed, causing newly built modules to also be uncompressed, even if the kernel supports compression.

Based on 07de9295c42e2ed58c25d8598e17745c5aae3c29, which already added this new behavior to run_test.sh.

Script demonstrating the issue:
```
#!/bin/bash

echo //// Compression settings from /lib/modules/$(uname -r)/build/.config ////
grep -e "CONFIG_MODULE_COMPRESS" /lib/modules/$(uname -r)/build/.config
echo

install_tree=/lib/modules

#### Snippet taken from dkms.in:234 ####
# Figure out the correct module suffix for the kernel we are currently
# dealing with, which may or may not be the currently installed kernel.
set_module_suffix()
{
    # $1 = the kernel to base the module_suffix on
    kernel_test="${1:-$(uname -r)}"
    module_uncompressed_suffix=".ko"
    grep -q '\.gz:' $install_tree/$kernel_test/modules.dep 2>/dev/null && module_compressed_suffix=".gz"
    grep -q '\.xz:' $install_tree/$kernel_test/modules.dep 2>/dev/null && module_compressed_suffix=".xz"
    grep -q '\.zst:' $install_tree/$kernel_test/modules.dep 2>/dev/null && module_compressed_suffix=".zst"
    module_suffix="$module_uncompressed_suffix$module_compressed_suffix"
}

echo //// Full module file extension determined by the current method ////
set_module_suffix $(uname -r)
echo $module_suffix
echo

KERNEL_VER=$(uname -r)

#### Snippet taken from run_test.sh:230 ####
mod_compression_ext=
kernel_config="/lib/modules/${KERNEL_VER}/build/.config"
if [ -f "${kernel_config}" ]; then
    if grep -q "^CONFIG_MODULE_COMPRESS_NONE=y" "${kernel_config}" ; then
        mod_compression_ext=
    elif grep -q "^CONFIG_MODULE_COMPRESS_GZIP=y" "${kernel_config}" ; then
        mod_compression_ext=.gz
    elif grep -q "^CONFIG_MODULE_COMPRESS_XZ=y" "${kernel_config}" ; then
        mod_compression_ext=.xz
    elif grep -q "^CONFIG_MODULE_COMPRESS_ZSTD=y" "${kernel_config}" ; then
        mod_compression_ext=.zst
    fi
fi

module_uncompressed_suffix=".ko"

echo //// Full module file extension expected by the currently failing test ////
echo "$module_uncompressed_suffix$mod_compression_ext"
```

Output on my machine:
```
user@host:~$ ./example.sh
//// Compression settings from /lib/modules/6.5.0-17-generic/build/.config ////
# CONFIG_MODULE_COMPRESS_NONE is not set
# CONFIG_MODULE_COMPRESS_GZIP is not set
# CONFIG_MODULE_COMPRESS_XZ is not set
CONFIG_MODULE_COMPRESS_ZSTD=y

//// Full module file extension determined by the current method ////
.ko

//// Full module file extension expected by the currently failing test ////
.ko.zst
```

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -238,9 +238,9 @@ set_module_suffix()
     # $1 = the kernel to base the module_suffix on
     kernel_test="${1:-$(uname -r)}"
     module_uncompressed_suffix=".ko"
-    grep -q '\.gz:' $install_tree/$kernel_test/modules.dep 2>/dev/null && module_compressed_suffix=".gz"
-    grep -q '\.xz:' $install_tree/$kernel_test/modules.dep 2>/dev/null && module_compressed_suffix=".xz"
-    grep -q '\.zst:' $install_tree/$kernel_test/modules.dep 2>/dev/null && module_compressed_suffix=".zst"
+    grep -q '^CONFIG_MODULE_COMPRESS_GZIP=y' $kernel_config 2>/dev/null && module_compressed_suffix=".gz"
+    grep -q '^CONFIG_MODULE_COMPRESS_XZ=y' $kernel_config 2>/dev/null && module_compressed_suffix=".xz"
+    grep -q '^CONFIG_MODULE_COMPRESS_ZSTD=y' $kernel_config 2>/dev/null && module_compressed_suffix=".zst"
     module_suffix="$module_uncompressed_suffix$module_compressed_suffix"
 }
 
@@ -1041,8 +1041,8 @@ prepare_build()
         $"Directory $base_dir already exists. Use the dkms remove function before trying to build again."
 
     # Read the conf file
-    set_module_suffix "$kernelver"
     read_conf_or_die "$kernelver" "$arch"
+    set_module_suffix "$kernelver"
 
     # Error out if build_exclude is set
     [[ $build_exclude ]] && diewarn 77 \
@@ -1788,8 +1788,8 @@ do_status_weak()
 # Spit out all the extra status information that people running DKMS are
 # interested in, but that the DKMS internals do not usually care about.
 module_status_built_extra() (
-    set_module_suffix "$3"
     read_conf "$3" "$4" "$dkms_tree/$1/$2/source/dkms.conf" 2>/dev/null
+    set_module_suffix "$3"
     [[ -d $dkms_tree/$1/original_module/$3/$4 ]] && echo -n " (original_module exists)"
     for ((count=0; count < ${#dest_module_name[@]}; count++)); do
         tree_mod=$(compressed_or_uncompressed "$dkms_tree/$1/$2/$3/$4/module" "${dest_module_name[$count]}")


masonlherndon:
I only tested by building/installing/removing a custom driver from work, and verifying that compression was properly detected and applied to the module.

I'm also not 1000% sure that the `.config` file is guaranteed to exist during every use-case where `set_module_suffix()` is called, so I've very open to suggestions. :)

masonlherndon:
> Generally I'm in favour. More so, we should probably remove this in the long run - see #319 for more.
> 
> I'm inclined to merge this, just a couple of questions. In you test/example above:
> 
>     * what distro are you using?
> 
>     * was the linux package installed?
> 
>     * does the linux package include "modules.dep", if not - what's the call chain/trigger for `depmod`?
> 
> 
> And to answer your question:
> 
> > I'm also not 1000% sure that the .config file is guaranteed to exist during every use-case where set_module_suffix() is called, which is maybe why we're reading modules.dep in the first place??
> 
> I _think_ that in the early days it was not guaranteed - with distro packaging varying greatly. These days, it's a hard-requirement for dkms and all the distros in our CI have it.

The output I showed was from a Ubuntu 22.04 machine. The relevant installed packages for this kernel version are `linux-image-6.5.0-17-generic`, `linux-headers-6.5.0-17-generic`, and `linux-modules-6.5.0-17-generic`, which all came from the meta-package `linux-generic-hwe-22.04`. I downloaded the .deb for `linux-modules-6.5.0-17-generic` to check if it came with a default `modules.dep`, however it seems to only have `modules.builtin` and `modules.order`. I don't know quite what you mean or how to find out what the call chain/trigger for `depmod` is.

masonlherndon:
> On second thought, I don't think we can do this _yet_.
> 
> Namely: Some distributions deliberately have a miss-match between the config and actual files shipped. See the run_tests.sh snippet for example
> 
> https://github.com/dell/dkms/blob/c1c2986a935fe7a326a1418183e6424906a7be4b/run_test.sh#L246-L266
> 
> So as-is this will cause a regression and there won't be an easy way for user/distro to reinstate the original behaviour. There's some ideas to have this as config toggle - see #328.
> 
> If you'd like to add a config for compression override type, then we can have this on top.

Doesn't the use of the `compressed_or_uncompressed()` function make the `.config` method of setting `module_compressed_suffix` a non-issue though? When a distro ships a `.ko` for a kernel that also supports compression, then `compressed_or_uncompressed()` will just pick the `.ko` path since the `.ko.<blah>` path won't be found.

masonlherndon:
> If you'd like to add a config for compression override type, then we can have this on top.
Are you suggesting like a `--compression-type` CLI option?

masonlherndon:
So, should I close this PR? Leave it open to maybe get merged in the distant future?

I don't think I'm gonna have time to implement a `module_compression` field anytime soon.

scaronni:
I'm sorry but I've merged something different than getting the compression from the `.config` file. On Fedora/RHEL (& derivatives) the kernel modules are not compressed in the configuration file but rather later in the build process:
```
$ cat /lib/modules/6.11.10-300.fc41.x86_64/config | grep MODULE_COMPRESS
CONFIG_MODULE_COMPRESS_NONE=y
# CONFIG_MODULE_COMPRESS_GZIP is not set
# CONFIG_MODULE_COMPRESS_XZ is not set
# CONFIG_MODULE_COMPRESS_ZSTD is not set
$ find /lib/modules/6.11.10-300.fc41.x86_64/ -name "*xz" | head -3
/lib/modules/6.11.10-300.fc41.x86_64/kernel/arch/x86/crypto/aegis128-aesni.ko.xz
/lib/modules/6.11.10-300.fc41.x86_64/kernel/arch/x86/crypto/blowfish-x86_64.ko.xz
/lib/modules/6.11.10-300.fc41.x86_64/kernel/arch/x86/crypto/camellia-aesni-avx-x86_64.ko.xz
```

On top of that, the kernel modules are not part of the kernel package, the base kernel package (`kernel-core`, which is used for minimal VMs, etc.) does not contain any. So the only way to get the module compression algorithm is through checking what's in the kernel tree.

I've changed the code and added some comment so the next person looking at this does not step into the same issue: https://github.com/dell/dkms/blob/master/dkms.in#L261-L274

Thanks.

PR #415, opened 2024-04-25
Make make_tarball vars writable to allow multiple kernel vers

elcabaye:
Without this change, when passing multiple -k to `dkms mktarball`, only the first one would be added to the tarball.


scaronni:
Closing as it has been posted again for review as the original proposer is MIA.

PR #417, opened 2024-05-21
dkms: fix building nvidia open kernel modules against clang and thin/full lto compiled kernel

### LINKED ISSUE #416: Failing to compile nvidia-open-dkms while using a ThinLTO/Clang kernel
```
dkms: 3.0.12-1
clang: 17.0.6-2
```

As there been some debate over this on TKG and NVIDIA github (like https://github.com/NVIDIA/open-gpu-kernel-modules/issues/405), I'm facing something different, but I don't know whether it's related to dkms or not, for example, when trying to install nvidia-open-dkms on Arch Linux, I see this:

```
strip: BFD (GNU Binutils) 2.42.0 assertion fail /usr/src/debug/binutils/binutils-gdb/bfd/elf.c:4131
strip: BFD (GNU Binutils) 2.42.0 assertion fail /usr/src/debug/binutils/binutils-gdb/bfd/elf.c:4131
...
strip: BFD (GNU Binutils) 2.42.0 assertion fail /usr/src/debug/binutils/binutils-gdb/bfd/elf.c:4131
strip: BFD (GNU Binutils) 2.42.0 assertion fail /usr/src/debug/binutils/binutils-gdb/bfd/elf.c:4131
strip: BFD (GNU Binutils) 2.42.0 assertion fail /usr/src/debug/binutils/binutils-gdb/bfd/elf.c:4131
strip: BFD (GNU Binutils) 2.42.0 assertion fail /usr/src/debug/binutils/binutils-gdb/bfd/elf.c:4131
...
```

(It will keep printing a bunch of those, it will still successfully install the nvidia-open-dkms besides that)

Also trying to boot the system with the borked driver won't work (it will hang on boot).

So I tried replacing the `/usr/bin/strip` for `/usr/bin/llvm-strip`, and those assertions disappears and the nvidia-open kernel module installs normally, though it still borked (it will hang on boot)

I think maybe the dkms script isn't setting the right `STRIP=/usr/bin/llvm-stip` environment variable or other Clang related stuff?

ltsdw:
Right now there's some issues when trying to compile open-gpu-kernel-modules against a kernel compiled using Clang and Thin/Full LTO.

Also the script will pick the wrong version of the `strip` command otherwise the module will fail to compile successfully:
```
strip: BFD (GNU Binutils) 2.42.0 assertion fail /usr/src/debug/binutils/binutils-gdb/bfd/elf.c:4131
strip: BFD (GNU Binutils) 2.42.0 assertion fail /usr/src/debug/binutils/binutils-gdb/bfd/elf.c:4131
strip: BFD (GNU Binutils) 2.42.0 assertion fail /usr/src/debug/binutils/binutils-gdb/bfd/elf.c:4131
...
```
Is also necessary to set the `OBJCOPY` environment variable, otherwise even though the modules will compile the system will refuse to boot on it.

Fixes #416.

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -625,17 +625,11 @@ read_conf()
     # Check if clang was used to compile or lld was used to link the kernel.
     if [[ -e $kernel_source_dir/vmlinux ]]; then
       if  readelf -p .comment $kernel_source_dir/vmlinux | grep -q clang; then
-        make_command="${make_command} CC=clang"
-      fi
-      if  readelf -p .comment $kernel_source_dir/vmlinux | grep -q LLD; then
-        make_command="${make_command} LD=ld.lld"
+        make_command="${make_command} LLVM=1"
       fi
     elif [[ -e "${kernel_config}" ]]; then
       if grep -q CONFIG_CC_IS_CLANG=y "${kernel_config}"; then
-        make_command="${make_command} CC=clang"
-      fi
-      if grep -q CONFIG_LD_IS_LLD=y "${kernel_config}"; then
-        make_command="${make_command} LD=ld.lld"
+        make_command="${make_command} LLVM=1"
       fi
     fi
 
@@ -1150,7 +1144,11 @@ actual_build()
         local built_module="$the_module$module_uncompressed_suffix"
         local compressed_module="$the_module$module_suffix"
 
-        [[ ${strip[$count]} != no ]] && strip -g "$built_module"
+        if [[ ${strip[$count]} != no ]] && [[ ${CC} == "clang" ]]; then
+            llvm-strip -g "$built_module"
+        elif [[ ${strip[$count]} != no ]]; then
+            strip -g "$built_module"
+        fi
 
         if (( do_signing )); then
             echo "Signing module $built_module"


ltsdw:
Thank you @evelikov .

Besides all the external modules I have compiling and installing normally on my system.
I ran the available dkms tests, this were the results:

```
Using kernel 6.9.1-2/x86_64
Checking module compression ...
config: CONFIG_MODULE_COMPRESS_ZSTD=y
files: /lib/modules/6.9.1-2/kernel/arch/x86/crypto/aesni-intel.ko.zst
Expected extension: .zst
Preparing a clean test environment
Test framework file hijacking
Adding the test module by version (expected error)
Adding the test module by directory
Adding the test module again (expected error)
Adding the test module by version (expected error)
Building the test module
--- test_cmd_expected_output.log	2024-06-04 16:42:18.060006565 -0300
+++ test_cmd_output.log	2024-06-04 16:42:19.721350689 -0300
@@ -1,6 +1,6 @@

 Building module:
 Cleaning build area...
-Building module(s)...
+make -j1 KERNELRELEASE=6.9.1-2 -C /usr/lib/modules/6.9.1-2/build M=/var/lib/dkms/dkms_test/1.0/build CC=clang OBJCOPY=llvm-objcopy LD=ld.lld...
 Signing module /var/lib/dkms/dkms_test/1.0/build/dkms_test.ko
 Cleaning build area...
Error: unexpected output from: dkms build -k 6.9.1-2 -m dkms_test -v 1.0
```

It builds and install the dkms_test normally but differs only by outputting the command used to build the module.

All seems fine.

ltsdw:
This is the line in the dkms.in file that outputs the command:

```
[[ $verbose ]] && echo -e "$1" || echo -en "$2..."
```

Exactly because $verbose is null, the second output is given echo -en "$2...".

ltsdw:
First, sorry about that, right now the arch linux package is sitting on version 3.0.12, I had to build 3.0.13 myself. Now the tests passes but with some caveats:

Tests output: [dkms_tests_output.txt](https://github.com/user-attachments/files/15570994/dkms_tests_output.txt)

Here are the changes I made to the run_test.sh (it isn't a suggestion to change to exact this, as somethings I had to hardcode because I couldn't get around it):

```patch
@@ -496,9 +496,9 @@
 echo 'Checking modinfo'
 run_with_expected_output sh -c "modinfo /lib/modules/${KERNEL_VER}/${expected_dest_loc}/dkms_test.ko${mod_compression_ext} | head -n 4" << EOF
 filename:       /lib/modules/${KERNEL_VER}/${expected_dest_loc}/dkms_test.ko${mod_compression_ext}
-version:        1.0
-description:    A Simple dkms test module
 license:        GPL
+description:    A Simple dkms test module
+version:        1.0
 EOF
 
 if (( NO_SIGNING_TOOL == 0 )); then
@@ -621,9 +621,9 @@
 echo 'Checking modinfo'
 run_with_expected_output sh -c "modinfo /lib/modules/${KERNEL_VER}/${expected_dest_loc}/dkms_test.ko${mod_compression_ext} | head -n 4" << EOF
 filename:       /lib/modules/${KERNEL_VER}/${expected_dest_loc}/dkms_test.ko${mod_compression_ext}
-version:        1.0
-description:    A Simple dkms test module
 license:        GPL
+description:    A Simple dkms test module
+version:        1.0
 EOF
 
 if (( NO_SIGNING_TOOL == 0 )); then
@@ -1419,7 +1419,7 @@
 Cleaning build area...
 Building module(s)...(bad exit status: 2)
 Failed command:
-make -j1 KERNELRELEASE=${KERNEL_VER} all
+make -j1 KERNELRELEASE=${KERNEL_VER} all CC=clang OBJCOPY=llvm-objcopy LD=ld.lld
 Error! Bad return status for module build on kernel: ${KERNEL_VER} (${KERNEL_ARCH})
 Consult /var/lib/dkms/dkms_failing_test/1.0/build/make.log for more information.
 dkms autoinstall on ${KERNEL_VER}/${KERNEL_ARCH} failed for dkms_failing_test(10)
@@ -1438,7 +1438,7 @@
 Cleaning build area...
 Building module(s)...(bad exit status: 2)
 Failed command:
-make -j1 KERNELRELEASE=${KERNEL_VER} all
+make -j1 KERNELRELEASE=${KERNEL_VER} all CC=clang OBJCOPY=llvm-objcopy LD=ld.lld
 Error! Bad return status for module build on kernel: ${KERNEL_VER} (${KERNEL_ARCH})
 Consult /var/lib/dkms/dkms_failing_test/1.0/build/make.log for more information.
 dkms autoinstall on ${KERNEL_VER}/${KERNEL_ARCH} failed for dkms_failing_test(10)
@@ -1587,7 +1587,7 @@
 Cleaning build area...
 Building module(s)...(bad exit status: 2)
 Failed command:
-make -j1 KERNELRELEASE=${KERNEL_VER} all
+make -j1 KERNELRELEASE=${KERNEL_VER} all CC=clang OBJCOPY=llvm-objcopy LD=ld.lld
 Error! Bad return status for module build on kernel: ${KERNEL_VER} (${KERNEL_ARCH})
 Consult /var/lib/dkms/dkms_failing_test/1.0/build/make.log for more information.
```

The explanation, my modinfo shows first the license, then the description and only then the version:

```
filename:       /lib/modules/6.9.1-2/kernel/drivers/hid/hid.ko.zst
author:         Andreas Gal
author:         Vojtech Pavlik
author:         Jiri Kosina
license:        GPL
vermagic:       6.9.1-2 SMP preempt mod_unload 
name:           hid
...
```

And about the variables CC, LD and OBJCOPY, right now the tests doesn't expect to encounter them, this isn't inherently from my PR, as we already been exposing the CC, and LD before (so the tests will fail on a clang compiled kernel even without my patch, so we may have to fix that. I even tried, but the said variables aren't exposed so we can't do something like `CC=${CC}` and so on, because CC isn't defined).

Other than that the tests passes normally.

ltsdw:
Yes, #422 fixes the failing tests for clang.

> Looking at the kernel documentation - I wonder if we shouldn't just use the LLVM=... variable all together. I'm not sure if tracking each and every variable, is going to be portable across multiple kernel releases.
> 
> [1] https://www.kernel.org/doc/html/latest/kbuild/llvm.html

Yes that would be better indeed. Would you like me to make this changes reflect this PR? We would still have to check for clang to use the correct strip though (even though LLVM=1 sets the STRIP envar).

ltsdw:
Sorry about that, I was trying not to create a bunch of commits while rebasing my local branch lol 😅 

> We do check the STRIP variable (array really) from the module's dkms.conf and make it an in-dkms variable. Even so we can technically honour the STRIP set by LLVM=1.

I didn't understand this part, for example, without this changes:

```patch
-        [[ ${strip[$count]} != no ]] && strip -g "$built_module"
+        if [[ ${strip[$count]} != no ]] && [[ ${CC} == "clang" ]]; then
+            llvm-strip -g "$built_module"
+        elif [[ ${strip[$count]} != no ]]; then
+            strip -g "$built_module"
+        fi
```

Even when using the LLVM=1 the wrong strip command will be picked. That's what I meant by still have to check for clang to use the write strip command.

> If you choose to do this (again up-to you), please update the man page.

If I understood the above incorrectly or you have any suggestion I'll be more than happy to implement and update this PR 😄

ltsdw:
OK, nvm, I understand now:

```sh
        case ${STRIP[$index]} in
            [nN]*)
                strip[$index]="no"
                ;;
            [yY]*)
                strip[$index]="yes"
                ;;
            '')
                strip[$index]=${strip[0]:-yes}
                ;;
        esac
```

The STRIP envar and the dkms.conf's STRIP array are two different things (got a little confused there). I get it now.

Well, if you don't like the approach of using the same logic as before (checking if the module should be stripped):

```sh
        if [[ ${strip[$count]} != no ]] && [[ ${CC} == "clang" ]]; then
            llvm-strip -g "$built_module"
        elif [[ ${strip[$count]} != no ]]; then
            strip -g "$built_module"
        fi
```

I can think of something else, but really I don't find it deemed necessary. The result would be the same, if the STRIP of dkms.conf is set "yes" the modules will be stripped, if set to "no", no stripping, and if empty the modules will be stripped, all paths leads to the right strip command being used and the LLVM=1 STRIP being honored no matter what.

Now, if I understood all wrong and by "honoring the STRIP set by LLVM=1" you meant that if clang is being used,  automatically we will use LLVM=1and inherently STRIP=llvm-strip will be set, meaning that the modules should be stripped by llvm-strip no matter what, I don't think it's a good idea, having control of when to strip modules in dkms.conf is better. (basically the way it's right now).

ptr1337:
@evelikov Would be cool, if you could review this in the near future.
560 Driver will use the open modules as default and this fixed an outstanding issue with them.

ltsdw:
@evelikov friendly ping 😅

ptr1337:
Were using this now too on CachyOS as default, mainly due the nvidia-open-dkms issue. Has been used for 1 months and no issues reported.

xuanruiqi:
This patch certainly works for me.
I think it should be fixed asap as lto-ed kernels are quite unusable with dkms without it

PR #419, opened 2024-06-03
dkms: include modules.weakdep in cleanup

liskin:
depmod now generates an additional file that needs to be cleaned.

See https://git.kernel.org/pub/scm/utils/kernel/kmod/kmod.git/commit/?id=05828b4a6e9327a63ef94df544a042b5e9ce4fe7


liskin:
It's already in Debian testing and unstable, which is where I noticed it 🙂

PR #428, opened 2024-07-07
Tests update

scaronni:
- CentOS 7 is EOL
- Ubuntu 18.04 is EOL, and extra updates up to 2028 are only available with an Ubuntu Pro subscription and the "`pro`" tool.
- Do not print dots on long operations when doing tests. I've found the parsing of the logs / removal of the dots very unreliable. This makes sure the extra progressing dots are not printed **only** during tests.
- Drop Ubuntu VM test as it's inconsistent with the output (hit in https://github.com/dell/dkms/pull/425#issuecomment-2217925669).


scaronni:
@evelikov or @xuzhen any chance for a review and merge?

scaronni:
> Do you have any references about the unreliable dot-dot-dot? Don't recall seeing any issues, although I haven't paid as much attention as I should have.

It's linked in the other MR above, just on the VM the container output is not consistent with the VM. Note that I also tested with a CentOS Stream 9 VM and there the output is consistent with the AlmaLinux 9 container.
 
> If it's not an isolated case, let's cross reference the yml in the dkms.in. So it's a bit harder to update one and not the other in the future.

Sorry not sure I understand, what do you mean?

> Please copy/paste the nice summary from the PR into the respective commits.

Done.

scaronni:
Btw, this is blocking the other pull requests as tests always fail. Thanks.

.github/workflows/tests.yml
```diff
@@ -104,7 +102,9 @@ jobs:
         apt-get install -qy gcc make linux-headers-generic linux-image-generic openssl shim-signed
 
     - name: Install dkms
-      run: make install
+      run: |
+        sed -i -e '/echo -en "."/d' dkms.in
```
zeroepoch, line 106: @scaronni, why not just remove these echo lines from `dkms.in` directly? Is the goal to make this change only when testing on GitHub, but not when run locally?

.github/workflows/tests.yml
```diff
@@ -143,38 +143,3 @@ jobs:
         else
             ./run_test.sh
         fi
-
-  test-vm:
```
zeroepoch, line 147: @scaronni you probably already know you can't really test kernel modules inside a container. This VM testing portion goes one step further than the container portion and appears to actually install the kernel modules and load them. IMO without this VM step it drops the test coverage in a pretty significant way.

zeroepoch: Ubuntu 20.04 will go EOL April next year so maybe it makes sense to try updating to 24.04? Or is the general strategy here to test older distro and then you'd expect newer distros to also work (somewhat by induction)?

run_test.sh
```diff
@@ -165,8 +165,6 @@ run_status_with_expected_output() {
 generalize_expected_output() {
     local output_log=$1
 
-    # "depmod..." lines can have multiple points. Replace them, to be able to compare
-    sed -i 's/\([^.]\)\.\.\.\.*$/\1.../' ${output_log}
```
zeroepoch, line 169: I have a different suggestion here that could fix the Ubuntu VM issue and in general be more robust to random extra newlines. What if we just remove all empty lines and then remove the extra newlines in the expected output?

```
# Remove lines containing only whitespace to normalize newline differences
sed -i '/^\s*$/d' ${output_log}
```

.github/workflows/tests.yml
scaronni: Correct, only in test.

This is to remove the printing of "progressing dots", i.e. every 3 seconds a new dot gets printed until the task is completed. Depending on the environment, the length of the dot line can of course be different. But when running it interactively on a system it's a useful indicator to check the status of the task, if it's stuck or not.

For example the Nvidia driver prints a very long line of dots when using the open source module tarball.

.github/workflows/tests.yml
scaronni: Unless you have the hardware for the modules available to the VM, even loading the modules in the VM does not bring anything. I think we should test that DKMS is just functionally correct.

scaronni: No we don't have any strategy for the VMs, it's just that matching the comment above, I don't see the value of having both the container and the VM.

run_test.sh
scaronni: We could, but then I don't think we will match the intended result in a merge request (for example) to... change the output.

zeroepoch: I feel like there is some reason the original author created a `dkms_test` module and added these tests. IMO if we can keep the test working that would be better, but as primarily a user of DKMS and only recently reviewing the code I don't think I would be able to say I have any strong opinions to keeping or removing the test. I'll leave that judgement best to @scaronni.

zeroepoch (approved):

scaronni:
Thanks!

Issue #430, opened 2024-07-11
Ship `kernel-install` snippet to `/usr/` instead of `/etc/`


behrmann:
Currently the `40-dkms.install` snippet is shipped to `$(DESTDIR)$(KCONF)/install.d/40-dkms.install`, but `KCONF` defaults to `/etc/kernel`. Unlike the `postinst.d` and `prerm.d` machinery `kernel-install` supports to have these files shipped in `/usr/`, which allows administrators to add superseding configuration in `/etc/`. It would therefore be helpful if the file could be shipped in `/usr/` instead.

bluca:
https://github.com/dell/dkms/pull/431

bluca:
related to https://github.com/dell/dkms/issues/343

Issue #343, opened 2023-09-21
Add support for stateless systems


evelikov:
As covered in https://0pointer.de/blog/projects/stateless.html.

The Tl:Dr: is we want a) the default config to be in /usr/..., where the sysadmin can either a) override the respectively named files or extend the config file snippets in /etc.

bluca:
This is only partially fixed, there are still some files shipped in etc

scaronni:
It got automatically closed when I closed https://github.com/dell/dkms/pull/431. Reopening...

PR #431, opened 2024-07-11
Install kernel-install drop-in to /usr/lib instead of /etc

bluca:
kernel-install supports picking configuration file from /usr and that's where vendor-defaults should ship to. /etc is for local modifications.

Partially fixes https://github.com/dell/dkms/issues/343
Fixes https://github.com/dell/dkms/issues/430


scaronni:
I'll make a test on Red Hat systems and merge if it works. Thanks!

PR #435, opened 2024-08-19
Move `local` to one var per line and init on seperate line

jcpunk:
This should resolve inconsistent results from shellcheck checks SC2178/SC2128

The tests pass for me locally, but extra careful eyes would be nice.

In theory this gets us closer to consistent coding practices.  I also renamed the variable `$mv` to `$modv` where I saw it as typos there may lead to confusion with the `mv` command.

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -77,8 +77,9 @@ _check_kernel_dir() {
 
 on_exit()
 {
-    local exitcode_on_exit=$?
-    local j="$(jobs -p)"
+    local exitcode_on_exit=$? # must be one line or $? captures rc from `local`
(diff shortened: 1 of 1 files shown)


scaronni:
Can you please rebase? I'll take a look just after.
thanks.

jcpunk:
rebased

scaronni:
@jcpunk tests do not run correctly. Please make sure nothing is broken and if the output has changed, please update the tests accordingly.

jcpunk:
Sorry about that, I forgot to copy my changed code into `/usr/bin/dkms`.

PR #452, opened 2024-10-21
Update tests

scaronni:
- Drop Debian 11 (EOL).
- Drop Ubuntu 23.10 (EOL).
- Add Fedora 41.
- Add Ubuntu 24.10.
- Switch `runs-on` to Ubuntu 24.04.
- Drop Gentoo; takes 30 minutes just to start the tests. Also, I don't quite get the Gentoo+DKMS combo, you rebuild the kernel, so along with that you rebuild the modules anyway.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,19 +29,18 @@ jobs:
           - {name: "centos/centos", tag: "stream9", url: "quay.io/"}
           - {name: "debian", tag: "testing"}
           - {name: "debian", tag: "12"}
-          - {name: "debian", tag: "11"}
           - {name: "fedora/fedora", tag: "rawhide", url: "quay.io/"}
+          - {name: "fedora/fedora", tag: "41", url: "quay.io/"}
           - {name: "fedora/fedora", tag: "40", url: "quay.io/"}
           - {name: "fedora/fedora", tag: "39", url: "quay.io/"}
-          - {name: "gentoo/stage3", tag: "latest"}
           - {name: "opensuse/tumbleweed", tag: "latest", variant: "-default", url: "registry.opensuse.org/"}
           - {name: "opensuse/leap", tag: "15.5", variant: "-default", url: "registry.opensuse.org/"}
           - {name: "opensuse/leap", tag: "15.6", variant: "-default", url: "registry.opensuse.org/"}
+          - {name: "ubuntu", tag: "24.10"}
           - {name: "ubuntu", tag: "24.04"}
-          - {name: "ubuntu", tag: "23.10"}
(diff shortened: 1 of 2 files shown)


Nowa-Ammerlaan:
> * Also, I don't quite get the Gentoo+DKMS combo, you rebuild the kernel, so along with that you rebuild the modules anyway.

What do you mean?? Gentoo is no different then any other distro in this regard (you know we also distribute a prebuilt kernel right?), out-of-tree kernel modules always need rebuilding on kernel updates. And not every module is packaged in the repository, that is where dkms comes in. I use it on my system to manage kernel modules that are only distributed in rpm/deb form. And I package this in our repository because I know others use it as well.

Frankly I am annoyed that you just drop this without even CC'ing me about it and that I have to find out when there is a new release and I can't find the results for the Gentoo tests. This test suite does not really run locally very well (it requires quite extensive write permissions) so I have been relying on the docker tests here to ensure that it works for us.

Please revert this, if there is a problem with the Gentoo tests then I will fix it (as I have in the past).

scaronni:
> Frankly I am annoyed that you just drop this without even CC'ing me about it and that I have to find out when there is a new release and I can't find the results for the Gentoo tests. This test suite does not really run locally very well (it requires quite extensive write permissions) so I have been relying on the docker tests here to ensure that it works for us.

First of all, don't expect me to contact you because I don't even know who you are.

I can ask you the same question, why didn't you fix the bug that has been opened for quite some time and fixed the gentoo tests that failed for months (mostly due to the readelf output)?

> Please revert this, if there is a problem with the Gentoo tests then I will fix it (as I have in the past). Or fix the tests properly so they can actually run in an unprivileged environment and then I can get them to automatically run from inside the package manager.

I'm not reverting anything, since you feel so entitled don't expect me to fix the tests for you. If you make a merge request that fixes both (test length and proper test output parsing) I'll be happy to merge.

Nowa-Ammerlaan:
> If you make a merge request that fixes both (test length and proper test output parsing) I'll be happy to merge.

PR is open here: https://github.com/dell/dkms/pull/454 

Note that the complaint about the missing `.comment` section was a real issue that needed fixing, not some quirk to be handled in test output parsing. I adjusted the code here to properly handled this case without making unneeded noise (after all, the `.comment` section is optional). And the next iteration of the prebuilt Gentoo kernel will exclude `vmlinux` from stripping by default, meaning it should then again have a `.comment` section. This I think makes sense considering the purpose of this file, no need to have users who want to use it go through the extra manual step of disabling stripping.

Also added several optimizations for speed, tests now complete successfully after ~4 minutes.

scaronni:
Awesome, thank you!

PR #454, opened 2024-10-23
Fix issue with stripped vmlinux and re-introduce Gentoo tests

Nowa-Ammerlaan:
- silence complaints from `readelf` if, for example, the `.comment` section is missing from `vmlinux`. If readelf fails then we should always just continue with the second `.config` based check.
- re-introduce the tests for Gentoo with more optimizations for speed.

[EDIT] Gentoo tests now complete successfully after 4 minutes, I'd say that is quite fast.


scaronni:
Perfect, than you!

PR #469, opened 2024-12-15
move kernel_prerm.d logic into dkms

anbe42:
this adds two new actions `kernel_postinst` (a wrapper for autoinstall) and `kernel_prerm` that are intended to be called from `kernel_*.d/dkms`  to move the prerm logic into dkms as well

I haven't done proper testing of the integration bits (the kernel_*.d changes) yet in my Debian dkms torture chroots, yet, and I won't be able to do them for the redhat side

naming etc. is open for discussion

### DIFF
--- a/debian_kernel_postinst.d.in
+++ b/debian_kernel_postinst.d.in
@@ -3,27 +3,6 @@
 # We're passed the version of the kernel being installed
 inst_kern=$1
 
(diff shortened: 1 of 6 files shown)


scaronni:
Thanks! Will perform all the necessary tests on Red Hat / SUSE systems this week.

PR #474, opened 2025-01-14
Ensure autoinstall reads /etc/dkms/$module.conf

bodgerer:
"dkms autoinstall" does not obey `BUILD_DEPENDS` (and presumably `AUTOINSTALL`) directives when placed in `/etc/dkms/$module.conf`, causing unexpected ordering and build failures.

The `read_conf` family of routines rely upon variables `module` and `module_version` (possibly others?) being already suitably set. Arguably they shouldn't, but this is a minimal change that corrects autoinstall's `BUILD_DEPENDS` processing.

There may be other places where this is needed.

Tested against dkms 3.1.4

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -2532,7 +2532,7 @@ autoinstall() {
             continue
         fi
         # If the module does not want to be autoinstalled, skip it.
-        read_conf_or_die "$kernelver" "$arch" "$dkms_tree/$m/$v/source/dkms.conf"
+        module=$m module_version=$v read_conf_or_die "$kernelver" "$arch" "$dkms_tree/$m/$v/source/dkms.conf"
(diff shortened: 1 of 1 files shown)


scaronni:
I think these are the only two variables required, as the function needs to run while iterating over `$m` + `$v`. It's also the only case where that function is called in a loop.

Thanks for the merge request!

bodgerer:
Hi Simone,

No problem : )

There may be another case lurking - I think I saw some attempts to read 
/etc/dkms/.conf from the "while read status mvka; do" loop, but didn't 
figure out where it was coming from

Best,

Mark

anbe42:
@bac0n Please file a new issue or PR, otherwise this comment on a closed PR tends to get forgotten ... Thanks!

Issue #455, opened 2024-10-31
Use faster zstd compression settings


skwerlman:
Currently, DKMS uses compression level 19 when ZSTD is selected as the compression method. However, this does not provide a significant benefit in terms of compressed size when compared to lower levels, especially when considering the time it takes to compress them:

As an example, here are the compression stats for `nvidia-dkms` on Arch Linux for compression levels `-19` and `-1`.

with `-19`:
```
 ╭─sk@a5 in ~ 
 ╰─λ zstd -l /var/lib/dkms/nvidia/560.35.03/6.11.5-zen1-1.1-zen/x86_64/module/nvidia*.zst  
Frames  Skips  Compressed  Uncompressed  Ratio  Check  Filename
     1      0    58.1 KiB       289 KiB  4.976  XXH64  /var/lib/dkms/nvidia/560.35.03/6.11.5-zen1-1.1-zen/x86_64/module/nvidia-drm.ko.zst
     1      0    49.7 MiB      96.8 MiB  1.947  XXH64  /var/lib/dkms/nvidia/560.35.03/6.11.5-zen1-1.1-zen/x86_64/module/nvidia.ko.zst
     1      0     685 KiB      2.72 MiB  4.072  XXH64  /var/lib/dkms/nvidia/560.35.03/6.11.5-zen1-1.1-zen/x86_64/module/nvidia-modeset.ko.zst
     1      0    2.23 KiB      8.54 KiB  3.824  XXH64  /var/lib/dkms/nvidia/560.35.03/6.11.5-zen1-1.1-zen/x86_64/module/nvidia-peermem.ko.zst
     1      0     839 KiB      6.22 MiB  7.599  XXH64  /var/lib/dkms/nvidia/560.35.03/6.11.5-zen1-1.1-zen/x86_64/module/nvidia-uvm.ko.zst
----------------------------------------------------------------- 
     5      0    51.2 MiB       106 MiB  2.068  XXH64  5 files
```

with `-1`:
```
 ╭─sk@a5 in ~ 
 ╰─λ zstd -l /var/lib/dkms/nvidia/560.35.03/6.11.5-zen1-1.1-zen/x86_64/module/nvidia*.zst
Frames  Skips  Compressed  Uncompressed  Ratio  Check  Filename
     1      0    73.4 KiB       289 KiB  3.940  XXH64  /var/lib/dkms/nvidia/560.35.03/6.11.5-zen1-1.1-zen/x86_64/module/nvidia-drm.ko.zst
[...]

Issue #386, opened 2024-01-25
dkms built modules may be compressed in a way unsupported by the kernel


anbe42:
from https://bugs.debian.org/1061427

dkms compresses some of the packages it builds. I guess this depends on the
kernel config that it builds for.
With xz compression, it compresses it in a way that's not compatible with
kernel's xz. The difference seems to be the CRC:

> file kernel/drivers/misc/isl29020.ko.xz: XZ compressed data, checksum CRC32
- mainline driver, works
> file updates/dkms/ddcci.ko.xz: XZ compressed data, checksum CRC64
- dkms driver, fails to load

anbe42:
this is about xz not zstd

the relevant option here should be `--check=crc32` 

but then again, why can't module compression be delegated to kbuild? the kernel probably knows best how it wants its modules ... does dkms really have to reinvent that?

anbe42:
in the Linux git master as of today:
```
$ grep -r -e --check=crc .
./Documentation/staging/xz.rst:--check=crc32.
./Documentation/staging/xz.rst: xz --check=crc32 --lzma2=dict=512KiB inputfile
./arch/powerpc/boot/wrapper:        xz --check=crc32 -f -6 "$vmz.$$"
./scripts/Makefile.modinst:      cmd_xz = $(XZ) --check=crc32 --lzma2=dict=1MiB -f $<
./scripts/Makefile.lib:      cmd_xzmisc = cat $(real-prereqs) | $(XZ) --check=crc32 --lzma2=dict=1MiB > $@
./scripts/xz_wrap.sh:exec $XZ --check=crc32 $BCJ --lzma2=$LZMA2OPTS,dict=32MiB
```

PR #476, opened 2025-01-17
Set compression levels to upstream defaults

C0rn3j:
Currently, the compression defaults are maxed out to maximize storage gains (except for xz which is set to default?), at the cost of taking minutes instead of seconds to compress.

There is previous conversation about this comparing zstd gains at max vs minimum - https://github.com/dell/dkms/issues/455

@skwerlman made a PR that that made the values configurable - https://github.com/dell/dkms/pull/456

This PR changes this configuration to upstream defaults, as suggested in the above discussion.

The suggestion also considered dropping the parameters altogether to just always adapt upstream defaults - it should be more user friendly for those that are trying to modify the configuration to have a set example.

This is likely to break setups where people are putting the modules on an undersized ESP (notably, when people reuse the 100MiB Windows ESP for /boot), so bumping the filesizes by a couple megabytes could tip the size over the max.  

Such people can either fix their ESP size, or more simply configure dkms to use the old values that sacrifice speed for space savings.

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -2691,9 +2691,9 @@ verbose=""
 symlink_modules=""
 
 # Set compression defaults
-compress_gzip_opts="-9"
-compress_xz_opts="--check=crc32 --lzma2=dict=1MiB"
-compress_zstd_opts="-q -T0 -19"
+compress_gzip_opts="-6"
+compress_xz_opts="--check=crc32 --lzma2=dict=1MiB -6"
+compress_zstd_opts="-q -T0 -3"
 
 # Check that we can write temporary files
 tmpfile=$(mktemp_or_die)
--- a/dkms_framework.conf.in
+++ b/dkms_framework.conf.in
@@ -51,6 +51,6 @@
 
 # Compression settings DKMS uses when compressing modules. The defaults are
 # tuned for maximum compression at the expense of speed when compressing.
-# compress_gzip_opts="-9"
-# compress_xz_opts="--check=crc32 --lzma2=dict=1MiB"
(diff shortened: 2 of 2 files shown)


christian-heusel:
Wouldn't it make more sense to omit these to also pick changed upstream defaults in the future?

C0rn3j:
> Wouldn't it make more sense to omit these to also pick changed upstream defaults in the future?

*The suggestion also considered dropping the parameters altogether to just always adapt upstream defaults - it should be more user friendly for those that are trying to modify the configuration to have a set example.*

C0rn3j:
Could do that for gzip, but I don't know how to cleanly do it for xz/zstd as they set more than just compression level value, and something like this looks mildly bad:

```bash
# compress_gzip_opts="-6"
compress_xz_opts="--check=crc32 --lzma2=dict=1MiB"
# compress_xz_opts="--check=crc32 --lzma2=dict=1MiB -6"
compress_zstd_opts="-q -T0"
# compress_zstd_opts="-q -T0 -3"
```

scaronni:
> EDIT: Or do you mean to leave the compression values only in the framework file?
> 
> EDIT2: Did that - won't set the defaults, but the commented out framework file has them as an example.

Yes, exactly. Thank you, I think it's fine.

skwerlman:
This is probably a good place to add a note about the kernel not supporting zstd's `--ultra` flag or xz's crc64 csum mode: https://github.com/dell/dkms/issues/386

scaronni:
Merging for now, but the proper solution would be https://github.com/dell/dkms/issues/319

PR #479, opened 2025-01-22
more autoinstall etc. tests

anbe42:
some reordering and enhancements for the testsuite

There are some more in the queue, but I need some CI changes for that but I have not enough experience with the github CI to fix that myself properly ...

I'm about to add some tests for `dkms_autoinstaller` (and maybe `dkms_common.postinst` later too) but that doesn't work on the CI since it only runs `make install` which skips these. It would need to run `make install_debian` or `make install_redhat` as appropriate for the current platform ...

(Actually I wanted to file this as a separate issue, but for some reason that didn't work ...)


anbe42:
> (Actually I wanted to file this as a separate issue, but for some reason that didn't work ...)

Creating issues works with `Ctrl+Return`, but not with clicking on the button ...

scaronni:
> I'm about to add some tests for dkms_autoinstaller (and maybe dkms_common.postinst later too) but that doesn't work on the CI since it only runs make install which skips these. It would need to run make install_debian or make install_redhat as appropriate for the current platform ...

I'll take care of this.

scaronni:
@anbe42 https://github.com/dell/dkms/pull/482

anbe42:
I rebased this PR on your `tests` branch, interleaved the previously omitted `dkms_autoinstaller` tests, tested and verified that they actually get executed.

I have a few more changes in the Debian package that should get upstreamed. I'll dig them out tomorrow and thereafter time should be right for a new release.

PR #482, opened 2025-01-22
Install specific targets during tests

scaronni:
- Run `make install-debian`, `make install-redhat` or just `make install` depending on the target.
- Drop sed replacement in `.github/workflows/tests.yml` that was referring to the previous way of printing dots.


anbe42:
yes, that would work

scaronni:
Deleted the branch by mistake...

Issue #488, opened 2025-02-05
/usr/bin/dkms: line 1910: continue: only meaningful in a 'for', 'while', or 'until' loop


Nowa-Ammerlaan:
`autoinstall` outputs a bunch of warnings about improper use of `continue`:
```
nowa-gentoo-laptop nowa # dkms autoinstall -k 6.13.1-gentoo-dist
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
/usr/bin/dkms: line 1910: continue: only meaningful in a `for', `while', or `until' loop
Sign command: /usr/src/linux/scripts/sign-file
Signing key: /root/kernel_key.pem
[...]

PR #493, opened 2025-02-12
Attempts for fixing #488

scaronni:
Tracking changes here for the above


scaronni:
> I would probably revert #490 and replace the stray `|| continue` with `|| return 1` instead

Ok!

scaronni:
Done.

scaronni:
@anbe42 please review, I think we can merge and close, based on last feedback in #488

anbe42 (approved):

PR #498, opened 2025-03-01
dkms.in: try to find Gentoo's module signing key and certificate

Nowa-Ammerlaan:
Just a little optimization for Gentoo so users don't have to configure the signing key and certificate in two places (with the package manager and in dkms `framework.conf`). Since we are now adding Gentoo cases for `$running_distribution` anyway, add those to some other places as well to avoid trying to call `rpm` unnecessarily.

See-also: https://github.com/gentoo/gentoo/pull/40704


scaronni:
That's cool, thank you!

PR #500, opened 2025-03-20
Fix disable autoinstall for a single module when AUTOINSTALL='no' is set

hongyuntw:
When we want to specifically disable the autoinstall functionality for a certain module, we can either remove the AUTOINSTALL-related settings from the module's dkms.conf file or set `AUTOINSTALL="no"`. However, the current `autoinstall` function does not handle this properly. In other words, even if a module has `AUTOINSTALL="no"`, it will still be automatically installed, which is not the expected behavior.


scaronni:
@hongyuntw please rebase the branch, we have test fixes already in main.

anbe42:
I have a slightly more generic solution queued ...

PR #506, opened 2025-03-31
Add codeSigning EKU to generated MOK

GabeNI:
Newer versions of openSUSE [require the `codeSigning` EKU on the module signing certificate](https://en.opensuse.org/openSUSE:UEFI#Signing_kernel_module).  This change will first attempt to set this value using the openssl `addext` argument.  If that fails, it will fall back to the previous openssl incanation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,8 @@ By default, DKMS generates a self signed certificate for signing modules at
 build time and signs every module that it builds before it gets compressed in
 the configured kernel compression mechanism of choice.
 
-This requires the `openssl` command to be present on the system.
+This requires version 1.1.1 or newer of the `openssl` command to be present on
+the system.
 
 Private key and certificate are auto generated the first time DKMS is run and
 placed in `/var/lib/dkms`. These certificate files can be pre-populated with
--- a/dkms.in
+++ b/dkms.in
@@ -1141,9 +1141,17 @@ prepare_mok()
             echo "openssl not found, can't generate key and certificate."
             return 1
         fi
+        # Newer openSUSE kernels require the codeSigning EKU
         openssl req -new -x509 -nodes -days 36500 -subj "/CN=DKMS module signing key" \
             -newkey rsa:2048 -keyout "$mok_signing_key" \
+            -addext "extendedKeyUsage=codeSigning" \
             -outform DER -out "$mok_certificate" > /dev/null 2>&1
+        if [[ ! -f ${mok_signing_key} ]]; then
+            # The addExt argument is supported on openssl 1.1.1 and newer.  Try again without it.
+            openssl req -new -x509 -nodes -days 36500 -subj "/CN=DKMS module signing key" \
+                -newkey rsa:2048 -keyout "$mok_signing_key" \
+                -outform DER -out "$mok_certificate" > /dev/null 2>&1
+        fi
         if [[ ! -f ${mok_signing_key} ]]; then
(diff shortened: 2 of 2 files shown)


scaronni:
Looks good to me. I can't find the requirement in the UEFI secure boot specification though. Do you know if it's a specific SUSE thing?

anbe42:
In that case lets avoid the fallback code. It can still be added later if the new option causes problems anywhere ...

Even if OpenSuSE was the first to expose this, I wouldn't mention it specifically in the comment to avoid having outdated questionable info there in the future. Sounds like a sane idea that might be picked up by others, too. Maybe just a generic `# Requires OpenSSL >= 1.1.1 for -addext`

Maybe mention the required version along the mentioning of openssl in README.md, too.

scaronni:
@GabeNI can you provide the necessary changes? thanks!

GabeNI:
> Looks good to me. I can't find the requirement in the UEFI secure boot specification though. Do you know if it's a specific SUSE thing?

SUSE is the only place I've encountered it so far.  I was surprised to encounter a message like "PKCS7: sinfo 1: The signer xxxxxxxx key is not CodeSigning" when trying to load a module signed with a key generated by DKMS.

After some brief spelunking, it looks like a [CONFIG_CHECK_CODESIGN_EKU option was added to the kernel](https://lkml2.uits.iu.edu/hypermail/linux/kernel/2105.2/10508.html) to assist with compliance to the NIAP Protection Profile for General Purpose Operating Systems.  Page 35 of [version 4.3 of that spec](https://www.niap-ccevs.org/protectionprofiles/469) states:

> * The OS shall validate the extendedKeyUsage field according to the
> following rules:
>    * Certificates used for trusted updates and executable code integrity
> verification shall have the Code Signing Purpose (id-kp 3 with OID
> 1.3.6.1.5.5.7.3.3) in the extendedKeyUsage field

As such, this check *could* end up on any distro's kernel in the future.

GabeNI:
> @GabeNI can you provide the necessary changes? thanks!

Done.  Thank you for the quick feedback.

anbe42 (approved):

scaronni:
> > Looks good to me. I can't find the requirement in the UEFI secure boot specification though. Do you know if it's a specific SUSE thing?
> 
> SUSE is the only place I've encountered it so far. I was surprised to encounter a message like "PKCS7: sinfo 1: The signer xxxxxxxx key is not CodeSigning" when trying to load a module signed with a key generated by DKMS.
> 
> After some brief spelunking, it looks like a [CONFIG_CHECK_CODESIGN_EKU option was added to the kernel](https://lkml2.uits.iu.edu/hypermail/linux/kernel/2105.2/10508.html) to assist with compliance to the NIAP Protection Profile for General Purpose Operating Systems. Page 35 of [version 4.3 of that spec](https://www.niap-ccevs.org/protectionprofiles/469) states:
> 
> > * The OS shall validate the extendedKeyUsage field according to the
> >   following rules:
> >   
> >   * Certificates used for trusted updates and executable code integrity
> >     verification shall have the Code Signing Purpose (id-kp 3 with OID
> >     1.3.6.1.5.5.7.3.3) in the extendedKeyUsage field
> 
> As such, this check _could_ end up on any distro's kernel in the future.

Thanks for the information, if no one beats me to it, I will make a merge request with this information written somewhere, I think it's good to have.

Otherwise merged.

PR #508, opened 2025-04-03
Update README.md

scaronni:
- Use Markdown's formatting for titles and paragraphs (`#`), this allows us to also have third level paragraphs.
- Drop 80 lines limit for the formatting, in the end it's just less readable when looking directly ad the unrendered Markdown file.
- Make sure every reference to the "program" is spelled capital (DKMS) and that every reference to the configuration file is marked as code (`dkms.conf`).
- Add paragraph about the CodeSigning key usage extension (relevant to https://github.com/dell/dkms/pull/506):
  - Regenerating it if needed on DKMS 3.1.7+.
  - Links to the specification.
  - Point out that the kernel config is out of tree, for now.
- Add paragraph about the UEFI specification for Secure Boot:
  - Add notes on the default key type.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,13 @@
-Dynamic Kernel Module System (DKMS)
-==
-This intention of this README is to explain how DKMS can be used in conjunction
-with tarballs which contain a dkms.conf file within them.
+# Dynamic Kernel Module System (DKMS)
+
+This intention of this README is to explain how DKMS can be used in conjunction with tarballs which contain a dkms.conf file within them.
 
 The DKMS project (and any updates) can be found at: https://github.com/dell/dkms
 
-Installation
---
 
-Installation is performed from the source directory with one of the following
-commands:
+## Installation
+
+Installation is performed from the source directory with one of the following commands:
 
 ```
 make install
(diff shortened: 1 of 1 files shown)


scaronni:
@GabeNI I've added the information from your merge request, thank you very much.

scaronni:
After this I will make a release as in Nvidia we're switching the SUSE builds (SLES and openSUSE Leap) to DKMS for version 575+, so I want to get this working properly.

anbe42:
What I found a good compromise for not wrapping lines is to have one sentence per line.

scaronni:
Good point, making sentences on a new line without a blank in between makes it more readable on the console but still keep sentences in the same block in the rendered file.

Amended.

scaronni:
View of the rendered file: https://github.com/dell/dkms/blob/sb-notes/README.md

anbe42 (approved):

anbe42:
> After this I will make a release as in Nvidia we're switching the SUSE builds (SLES and openSUSE Leap) to DKMS for version 575+, so I want to get this working properly.

Oh, I didn't know you were working at Nvidia  ;-) ... can I come with some packaging related questions to you?

If you ship/generate/... a dkms.conf please set BUILD_EXCLUSIVE_KERNEL_MIN as appropriate and probably use CLEAN=true as a backwards-compatible no-op.

PR #514, opened 2025-04-09
deprecate CLEAN

anbe42:
stop running clean commands in the build directory which will be removed immediately afterwards

emit deprecation message  for CLEAN unless the command is "true" for backwards compatibility with older dkms versions that still run the command (and where the default clean command would fail)

(this PR stacks on top of the previous one)


scaronni:
Hi @anbe42 , thank you for dedicating so much time on DKMS!

Can you please rebase? I merged the previous one as you mentioned, but there is a conflict. Thanks!

umlaeute:
is there some documentation on the impliciations of this, for oot kernel module developers?

that is: should I start dropping `CLEAN` from my dkms.conf *right now*? what will happen on distributions that (only) ship an older dkms?

[dkms versions in various distros](https://repology.org/project/dkms/versions)

(when) is it save to omit `CLEAN`? (my `dkms.conf` contains a `CLEAN="make KERNEL_DIR=${kernel_source_dir} clean"` which i guess is pretty standard)


Related https://github.com/v4l2loopback/v4l2loopback/issues/630

scaronni:
DKMS will just start printing that `CLEAN` is a deprecated variable and move on happily. Same thing happened with the other deprecated variables, like `REMAKE_INITRD`.

You can omit `CLEAN` since DKMS 3.1.8.

anbe42:
For a backwards-compatible CLEAN setting use `CLEAN=true` to turn this into a no-op for older dkms versions. CLEAN was useless from the very beginning, since immediately after cleaning the build tree the build tree will be deleted.

The value `true` will also suppress the deprecation message on 3.1.8+.

anbe42:
We should probably document this somewhere ...

Issue #499, opened 2025-03-04
reads wrong config file while checking BUILD_DEPENDS availability


anbe42:
While investigating a spurious autopkgtest failure in Debian (of the only(?) dkms package in Debian using BUILD_DEPENDS), I noticed that dkms claims the build dependency is not available after it has been successfully built and installed immediately before this check.

So far I haven't been able to reproduce that in run_tests.sh

It could be a recently introduced regression, since I think the test succeeded in the past. But it could also be a very fragile path that leads to an error here. If there is a second kernel, the failure seems to be only on the first.

What I noticed so far is:
- is_module_built gets 4 arguments and while checking availability of  a build-dependency, $1/$2 refer to the B-D being checked and differ from $module/$module_version. But when is_module_built calls read_conf_or_die, this happens with the implicit $module/$module_version parameters and a $conf override inherited from somewhere. As a result, some wrong module gets checked for being built returning a failure.
- If I prefix that command with module=$1 module_version=$2 and the correct dkms.conf (from the build-dependency) gets loaded, this modifies the internal state (BUILT_MODULE_NAME MAKE etc.) and the whole thing explodes during build.

scaronni:
@anbe42 do you want to tackle this in the short term or can I proceed with tagging and releasing 3.1.6?

scaronni:
Nevermind, I'll just tag the release, let's address it after. thanks.

anbe42:
that's a non-trivial issue and I have not yet fully understood what is happening here
for not it should not be a release blocker

anbe42:
Right now I cannot reproduce this issue. This is the second bug I noticed today that magically disappeared over the last month ... (the other was in an unrelated project) ;-)

scaronni:
It's because of all the cleanups you did 🥇

scaronni:
Let's close it, we can always reopen when that happens.

Issue #406, opened 2024-03-16
marking built/installed modules as rebuild-needed


anbe42:
When upgrading a linux-header (or linux-image) package without a kernel version change, we should rebuild all modules for that kernel. I'm a bit reluctant to `unbuild` all modules during preinst, I'd rather "mark" them for rebuilding during the next `autoinstall` run. Maybe this could be achieved by creating a flag file in the previously built module directory.
All actions that require the module to be built (build, install, autoinstall) should automatically `unbuild` the module if that flag file exists (which would cause the flag file to be deleted) before carrying out their operation.
I'm not sure whether we should change the `status` output if a module is in that state.
Ideally this should be implemented directly in dkms to be called from the scripts, e.g.
`dkms mark-for-rebuild -k <kernel> [-m/-v module/version]`
This should also operate on AUTOINSTALL=no modules.

Which also raises another question: shouldn't `kernel_prerm.d_dkms.in` act on 'built' modules, too, in addition to 'installed' ones?

PR #515, opened 2025-04-10
DRAFT: Add logic for rebuilding modules when dependencies change

scaronni:
Force the rebuild of modules when `BUILD_DEPENDS_REBUILD` is set and modules declared in `BUILD_DEPENDS` change.

When a module is updated, all modules that depend on it need to be automatically rebuilt in the correct order.

This is **all gated** through`BUILD_DEPENDS_REBUILD`, which should be set on a per installation basis, for example:
```
$ cat /etc/dkms/<module1>.conf
BUILD_DEPENDS=<module2>
BUILD_DEPENDS_REBUILD=yes
```

Whenever `<module1>` is built and `BUILD_DEPENDS` is set, the version of the dependent module is saved in:
```
/var/lib/dkms/<module>/<kernel_version>-<arch>/.dep_<dependency>
```

In case this changes, a new function `check_and_rebuild_dependent_modules()` that is called at the end of `do_install()` does the following:

Makes a first pass to collect all modules that need to be rebuilt by:
- Iterating through all modules
- Checking if they have `BUILD_DEPENDS` specified
- Using the previously stored dependency file to see if they need to be rebuilt
  - A missing file it's equal to a different version (so a rebuild)

Then it does a second pass to rebuild modules in dependency order by:
- Maintaining a list of already rebuilt modules
- For each module to rebuild, checking if all its dependencies have been rebuilt
- Only rebuilding modules whose dependencies are satisfied
- Failures:
  - Reports any modules that couldn't be rebuilt due to circular dependencies
  - Continues with the process even if some modules couldn't be rebuilt

If the user runs a `build/install` command with the `--force` parameter for a module where `BUILD_DEPENDS` and `BUILD_DEPENDS_REBUILD` are specified, the program will behave as before, bypassing the check and not saving the dependency in the file.

Sample run:
```
$ dkms status
2nd/1.0.0, 6.13.9-200.fc41.x86_64, x86_64: installed
1st/1.0.0, 6.13.9-200.fc41.x86_64, x86_64: installed

$ cat /var/lib/dkms/evdi/kernel-6.13.9-200.fc41.x86_64-x86_64/.dep_1st
1.0.0

# dkms remove -m 1st/1.0.0
Module 1st/1.0.0for kernel 6.13.9-200.fc41.x86_64 (x86_64):
Before uninstall, this module version was ACTIVE on this kernel.
Deleting /lib/modules/6.13.9-200.fc41.x86_64/extra/1st.ko.xz
Running depmod.... done.

Deleting module 1st/1.0.0 completely from the DKMS tree.

# dkms install -m 1st/2.0.0
Creating symlink /var/lib/dkms/1st/2.0.0/source -> /usr/src/1st-2.0.0

Sign command: /lib/modules/6.13.9-200.fc41.x86_64/build/scripts/sign-file
Signing key: /var/lib/dkms/mok.key
Public certificate (MOK): /var/lib/dkms/mok.pub

Building module(s)... done.
Signing module /var/lib/dkms/1st/2.0.0/build/1st.ko
Cleaning build area... done.
Installing /lib/modules/6.13.9-200.fc41.x86_64/extra/1st.ko.xz
Running depmod.... done.

Rebuilding module 2nd/1.0.0 due to updated dependencies

Module 2nd/1.0.0 for kernel 6.13.9-200.fc41.x86_64 (x86_64):
Before uninstall, this module version was ACTIVE on this kernel.
Deleting /lib/modules/6.13.9-200.fc41.x86_64/extra/2nd.ko.xz
Running depmod.... done.
Building module(s).... done.
Signing module /var/lib/dkms/2nd/1.0.0/build/2nd.ko
Cleaning build area... done.
Installing /lib/modules/6.13.9-200.fc41.x86_64/extra/2nd.ko.xz
Running depmod.... done.

$ cat /var/lib/dkms/evdi/kernel-6.13.9-200.fc41.x86_64-x86_64/.dep_1st
2.0.0
```

### DIFF
--- a/dkms.8.in
+++ b/dkms.8.in
@@ -739,8 +739,24 @@ This optional directive is an array that allows you to specify other modules as
 dependencies for your module. Each array element should be the
 .B PACKAGE_NAME
 of another module that is managed by dkms. Do not specify a version or
-architecture in the dependency. Note that this directive is only advisory;
-missing or broken dependencies cause non-fatal warnings.
+architecture in the dependency. A missing dependency requires the parameter
+.OP --force
+to still attempt the build of the module.
+.TP
+.B BUILD_DEPENDS_REBUILD=
+If this directive is set to
+.B yes
+then the module which has
+.B BUILD_DEPENDS
+being set, will get rebuilt every time a dependency changes version. This is usually provided in a specific
+.I /etc/dkms/<module>.conf
+on a per installation basis.
+
+To keep track of the dependency change, the version of the module that was used at build time is recorded in the file
+.I /var/lib/dkms/<module>/<kernel_version>-<arch>/.dep_<dependency>.
+As soon as the module version changes, this is compared and the dependent module is rebuilt. With the parameter
+.I --force
+the dependency is not tracked.
 .TP
 .B BUILD_EXCLUSIVE_KERNEL=
 This optional directive allows you to specify a regular expression which defines
--- a/dkms.in
+++ b/dkms.in
@@ -93,7 +93,7 @@ readonly dkms_conf_variables="PACKAGE_NAME
    built_module_name BUILT_MODULE_LOCATION built_module_location
    DEST_MODULE_NAME dest_module_name
    DEST_MODULE_LOCATION dest_module_location
-   STRIP strip AUTOINSTALL NO_WEAK_MODULES
+   STRIP strip AUTOINSTALL NO_WEAK_MODULES BUILD_DEPENDS_REBUILD
 
    CLEAN
    REMAKE_INITRD MODULES_CONF MODULES_CONF_OBSOLETES
@@ -582,8 +582,7 @@ safe_source() {
 # Source a dkms.conf file and perform appropriate postprocessing on it.
 # Do our best to not repeatedly source the same .conf file -- this can happen
 # when chaining module installation functions or autoinstalling.
-read_conf()
-{
+read_conf() {
     # $1 kernel version (required)
     # $2 arch (required)
     # $3 dkms.conf location (optional)
@@ -844,6 +843,21 @@ read_conf()
             ;;
         *)
             echo "dkms.conf: Error! Unsupported NO_WEAK_MODULES value '$NO_WEAK_MODULES'" >&2
+            return_value=1
+            ;;
+    esac
+
+    # Check for allowed BUILD_DEPENDS_REBUILD values: "yes", "no", ""
+    case "$BUILD_DEPENDS_REBUILD" in
+        "")
+            ;;
+        [Yy][Ee][Ss])
+            ;;
+        [Nn][Oo])
+            BUILD_DEPENDS_REBUILD=""
+            ;;
+        *)
+            echo "dkms.conf: Error! Unsupported BUILD_DEPENDS_REBUILD value '$BUILD_DEPENDS_REBUILD'" >&2
             return_value=1
             ;;
     esac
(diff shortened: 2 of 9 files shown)


scaronni:
@anbe42 before proceeding with implementing the tests and some stress testing with what I actually have to do, I would like your opinion on it.

anbe42:
I need a bit time to think about that ... so here are just some things that came into my mind while brainstorming about it.

First of all, that reminds me of #406, although I was looking at the orthogonal dimension: rebuild all modules (for a single kernel) if the kernel headers change (but $kernelver does not change).

The problem you want to solve effectively means we need to rebuild multiple modules for multiple kernels...

I have no idea what this means in the context of modules with multiple versions installed concurrently. That is a concept "we do not use" in Debian.

If something is going to introduce a circular dependency, can we detect and reject that at `dkms add` time?

scaronni:
> I have no idea what this means in the context of modules with multiple versions installed concurrently. That is a concept "we do not use" in Debian.

I'm not sure I understand. I fail to find a use case for modules with multiple versions installed concurrently, I also never saw it in the RHEL world.

The whole reasoning behind this, is to have the opportunity to rebuild `nvidia-peermem` once the Mellanox components change. We're planning to integrate the Mellanox repositories and bits into the CUDA repository and I need this functionality.

> If something is going to introduce a circular dependency, can we detect and reject that at `dkms add` time?

That's actually a good idea.

Since I don't see any particular objections on the general idea behind this, I'll fix the current tests, add more tests and then
add the logic & tests for the circular detection issue.

anbe42:
There is also #499 that still needs to be investigated ...

Let's for now assume we only have simple dependency graphs out in the wild (only short linear lists, no trees or dags), but still have the code ready for general graphs. For complicated graphs I'm afraid I'measily  able to come up with command orders where things break.

Right now when upgrading a module A, we unbuild that (which may make module B with BUILD_DEPENDS=A temporarily unbuildable), so if the package manager triggers building B, we may run into failures ...

When rebuilding modules, we should preserve the original state they were in: added/built/installed.

scaronni:
> How do you plan to model this dependency, since the driver (which contains -peermem) has no dependencies yet?  Currently it builds a a dummy -peermem module if the mellanox bits are missing from the kernel (and in Debian we have a patch which at least says so when attempting to load the module). Right now this sounds more like you need an "Enhances" (a package relationship supported by the .deb format, roughly a reverse Recommends). Or an OPTIONAL_BUILD_DEPENDS which does no harm when missing, but can cause rebuilds when available ...

I don't think we need all of this. What I was actually thinking is way simpler, in that particular case (which seems to be only the DGX stations from NVIDIA so far), it's up to the admin (or the DGX iso installer) to add the dependencies.

The point of the extra `dkms.conf` fragments is to customize the behavior for each site.

Example:
```
echo "BUILD_DEPENDS=<module>" > /etc/dkms/<module>.conf
```

Custom per-installation configuration is actually already happening for a few other modules that have different parameters for building depending on the context.

scaronni:
Another completely different approach would be an installation specific script that is pulled in by DKMS in one of the various stages. This way nothing changes in DKSM but you can still achieve the same,

I'll prototype that as well internally and see what's the best approach.

scaronni:
Added tests to check:
- Updates & rebuild of dependencies
- Checking the dependency after each state
- Checking that with `--force` it all behaves as if `BUILD_DEPENDS_REBUILD` is not set (dependencies ignore, dependency version not saved).

Now it will have a few days of testing in NVIDIA with various use cases before marking it ready for review.

scaronni:
Btw, I'm open to rename `BUILD_DEPENDS_REBUILD` to something else which is less of a mouthful, ideas accepted

scaronni:
Merging this. Added another safeguard when `--force` is passed and added also downgrade tests. Internal tests in NVIDIA on a DGX system with the Mellanox pile (https://linux.mellanox.com/public/repo/doca/DGX_latest_DOCA/ubuntu24.04/) are all good.

PR #516, opened 2025-04-11
dkms: propagate OBSOLETE_BY to BUILD_EXCLUSIVE_KERNEL_MAX

anbe42:
do not attempt to build modules that won't be installed

That's all what I have ready for now, #499 will take more time. If you don't have anything else in the queue, we should make a new release with the bugfixes we have in main so far.


scaronni:
> If you don't have anything else in the queue, we should make a new release with the bugfixes we have in main so far.

Doing a release now, so I can just provide that with a patch for https://github.com/dell/dkms/pull/515 for internal testing.

PR #526, opened 2025-05-10
Fix text in framework.conf

scaronni:
As part of https://github.com/dell/dkms/pull/522, we did a last minute change on the behavior of the `$kernel` variable, but the text in `framework.conf` was not updated accordingly.


amilonenv (approved):
LGTM

PR #531, opened 2025-05-29
dkms: continue on error

cshirron:
In the postinst script, if we hit an error building for one kernel, record the error and continue on to the next kernel. When done, exit with the value of any recorded error. This allows us to do a best-effort build for all installed kernels.


cshirron:
I have taken your suggestions and made the appropriate changes.

scaronni:
@cshirron sorry for the delay. Can you please make the necessary adjustments so that the pipelines pass? Thanks.

scaronni:
Sorry for the delay, forgot completely. Checking now.

Issue #471, opened 2024-12-19
Conditional built modules in dkms.conf


JDarnley:
Similar to #392 I am looking for a conditional listing of files in dkms.conf but in contrast to that (I think) I want it to be based on a kernel config option.  In my case I specifically want CONFIG_AUXILIARY_BUS.

I make a package using "dkms mkdeb" so I need it to be checked when apt tries to make dkms install it much later and elsewhere for every kernel that dkms tried to install for.

Is there an approved way to query this at "build" time?

ian-abbott:
I'm not recommending the following as a recommended solution, but it is what I do in [Comedi](https://github.com/Linux-Comedi/comedi) as a workaround.

The `dkms.conf` file gets sourced in as Bash source after some sanity checks, so it is possible to put the `BUILT_MODULE_NAME[n]`, `BUILT_MODULE_LOCATION[n]`, and `DEST_MODULE_LOCATION[n]` entries inside a Bash `if/then/fi` command. One problem is that the array indices need to be in an increasing sequence without gaps, but one way to do that is to use a local variable to hold the current index, and increment it.

For example:

```
local nn
nn=0
BUILT_MODULE_NAME[$nn]="foo1"
BUILT_MODULE_LOCATION[$nn]="ziggy/"
DEST_MODULE_LOCATION[$((nn++))]="/kernel/ziggy/"
BUILT_MODULE_NAME[$nn]="foo2"
BUILT_MODULE_LOCATION[$nn]="ziggy/"
DEST_MODULE_LOCATION[$((nn++))]="/kernel/ziggy/"
if grep -q '^CONFIG_AUXILIARY_BUS=[ym]' ${kernel_source_dir}/.config' 2>/dev/null; then
BUILT_MODULE_NAME[$nn]="foo-aux"
BUILT_MODULE_LOCATION[$nn]="ziggy/"
DEST_MODULE_LOCATION[$((nn++))]="/kernel/ziggy/"
fi
```

scaronni:
@ian-abbott any chance you could test? You just need to convert your if line in a directive with the module name.

Thanks!

ian-abbott:
> [@ian-abbott](https://github.com/ian-abbott) any chance you could test? You just need to convert your if line in a directive with the module name.

Some of the tests I use in https://github.com/Linux-Comedi/comedi/blob/master/dkms.conf.in are not quite so straightforward, such as this one:

```
# Modules that require set_fs() support in the kernel
[...]

PR #533, opened 2025-07-01
Add per-module conditionals

scaronni:
Add per-module `BUILD_EXCLUSIVE_*` directives in DKMS.

* Backward compatibility: global `BUILD_EXCLUSIVE` directives still work as before.
* Per-module support: new array-based directives `BUILD_EXCLUSIVE_KERNEL[#]`, `BUILD_EXCLUSIVE_KERNEL_MIN[#]`, `BUILD_EXCLUSIVE_KERNEL_MAX[#]`, `BUILD_EXCLUSIVE_ARCH[#]`, and `BUILD_EXCLUSIVE_CONFIG[#]`.
* Combined logic: both global and per-module directives are checked, and if any module fails either set of constraints, the build is excluded.

Implements https://github.com/dell/dkms/issues/471.

### DIFF
--- a/test/dkms_per_module_config_exclusive_test-1.0/dkms.conf
+++ b/test/dkms_per_module_config_exclusive_test-1.0/dkms.conf
@@ -1,0 +1,15 @@
+PACKAGE_NAME="dkms_per_module_config_exclusive_test"
+PACKAGE_VERSION="1.0"
+
+BUILT_MODULE_NAME[0]="dkms_per_module_config_exclusive_test_mod1"
+BUILT_MODULE_NAME[1]="dkms_per_module_config_exclusive_test_mod2"
+
+DEST_MODULE_LOCATION[0]="/updates/dkms"
+DEST_MODULE_LOCATION[1]="/updates/dkms"
+
+# Requires CONFIG_MODULES to be enabled and CONFIG_NONEXISTENT_CONFIG to be absent (should always match):
+BUILD_EXCLUSIVE_CONFIG[0]="CONFIG_MODULES !CONFIG_NONEXISTENT_CONFIG"
+# Requires CONFIG_NONEXISTENT_CONFIG to be enabled and CONFIG_MODULES to be abset (should never match):
+BUILD_EXCLUSIVE_CONFIG[1]="!CONFIG_MODULES CONFIG_NONEXISTENT_CONFIG"
+
+AUTOINSTALL="yes"
--- a/test/dkms_per_module_config_test-1.0/dkms.conf
+++ b/test/dkms_per_module_config_test-1.0/dkms.conf
@@ -1,0 +1,18 @@
+PACKAGE_NAME="dkms_per_module_config_test"
+PACKAGE_VERSION="1.0"
+
+BUILT_MODULE_NAME[0]="dkms_per_module_config_test_mod1"
+BUILT_MODULE_NAME[1]="dkms_per_module_config_test_mod2"
+BUILT_MODULE_NAME[2]="dkms_per_module_config_test_mod3"
+
+DEST_MODULE_LOCATION[0]="/updates/dkms"
+DEST_MODULE_LOCATION[1]="/updates/dkms"
+DEST_MODULE_LOCATION[2]="/updates/dkms"
+
+# Build only for kernel starting with 99 (should never match):
+BUILD_EXCLUSIVE_KERNEL[0]="^99\..*"
+# Build for any kernel (should always match) but only for architecture "none" (should never match):
+BUILD_EXCLUSIVE_KERNEL[1]=".*"
+BUILD_EXCLUSIVE_ARCH[1]="none"
+
+AUTOINSTALL="yes"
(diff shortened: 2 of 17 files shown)


scaronni:
If there are no objections I will merge this. Extensive tests have been added.

test/dkms_per_module_config_test-1.0/dkms.conf
```diff
@@ -0,0 +1,26 @@
...
+# Multiple modules with different per-module BUILD_EXCLUSIVE configurations
+BUILT_MODULE_NAME[0]="dkms_per_module_config_test_mod1"
+BUILT_MODULE_NAME[1]="dkms_per_module_config_test_mod2"
+BUILT_MODULE_NAME[2]="dkms_per_module_config_test_mod3"
+
+DEST_MODULE_LOCATION[0]="/updates/dkms"
+DEST_MODULE_LOCATION[1]="/updates/dkms"
+DEST_MODULE_LOCATION[2]="/updates/dkms"
+
+# Per-module kernel version exclusions
```
zeroepoch, line 13: I believe these conditions are inverted? Exclusive would mean only in "this particular" case, right? If not this usage is inconsistent with the test `test/dkms_per_module_config_exclusive_test-1.0/dkms.conf`, which matched my understanding as requiring a matching (exclusive to) config.

For the cases below:
- 0: Only build when kernel version is 99.x
- 1: Build when kernel version is anything
- 2: This is the same as 2?

test/dkms_per_module_config_test-1.0/dkms.conf
```diff
@@ -0,0 +1,26 @@
...
+# Module 0: exclude kernel versions starting with 99 (should never match)
+BUILD_EXCLUSIVE_KERNEL[0]="^99\..*"
+# Module 1: exclude all kernel versions (should always exclude)
+BUILD_EXCLUSIVE_KERNEL[1]=".*"
+# Module 2: no per-module restrictions
+
+# Per-module architecture exclusions
+# Module 0: no arch restrictions
+# Module 1: exclude current architecture (should exclude)
```
zeroepoch, line 22:
```suggestion
# Module 1: exclude all architectures (should exclude)
```

test/dkms_per_module_config_exclusive_test-1.0/dkms.conf
```diff
@@ -0,0 +1,17 @@
...
+BUILT_MODULE_NAME[1]="dkms_per_module_config_exclusive_test_mod2"
+
+DEST_MODULE_LOCATION[0]="/updates/dkms"
+DEST_MODULE_LOCATION[1]="/updates/dkms"
+
+# Per-module config restrictions
+# Module 0: requires CONFIG_MODULES to be enabled (should always match)
+BUILD_EXCLUSIVE_CONFIG[0]="CONFIG_MODULES"
+# Module 1: requires CONFIG_NONEXISTENT_CONFIG to be enabled (should never match)
+BUILD_EXCLUSIVE_CONFIG[1]="CONFIG_NONEXISTENT_CONFIG"
```
zeroepoch, line 15: You should also add cases for `!CONFIG_MODULES` and `!CONFIG_NONEXISTENT_CONFIG` to test the negative cases

scaronni: Good point, will add it.

test/dkms_per_module_config_test-1.0/dkms.conf
scaronni: I changed the text completely so it's more clear.

test/dkms_per_module_config_exclusive_test-1.0/dkms.conf
scaronni: Added both conditions in the same `BUILD_EXCLUSIVE_CONFIG`, so now there are two conditions for each directive (one true and one false).

test/dkms_per_module_config_test-1.0/dkms.conf
scaronni: Changed all comment text to make it clearer.

scaronni:
I've also removed one extra set of tests, they were too redundant.

zeroepoch (approved):

test/dkms_per_module_kernel_version_test-1.0/dkms.conf
```diff
@@ -0,0 +1,15 @@
...
+BUILT_MODULE_NAME[0]="dkms_per_module_kernel_version_test_old"
+BUILT_MODULE_NAME[1]="dkms_per_module_kernel_version_test_new"
+
+DEST_MODULE_LOCATION[0]="/updates/dkms"
+DEST_MODULE_LOCATION[1]="/updates/dkms"
+
+# Build only for kernels < 5.0 (should never match):
+BUILD_EXCLUSIVE_KERNEL_MAX[0]="4.99"
+# Build only for kernels >= 5.0 (should always match):
+BUILD_EXCLUSIVE_KERNEL_MIN[1]="5.0"
```
anbe42, line 13: How many modules is this test supposed to build?

anbe42 (commented):
I suspect that is not yet working as expected ...
Does (should?) the behavior change of you swap the modules and their conditions?

scaronni:
> does this work (and build mod_always):
> 
> BUILT_MODULE_NAME[0]="mod_never" BUILD_EXCLUSIVE_CONFIG[0]="CONFIG_NEVER" BUILT_MODULE_NAME[1]="mod_always"

Yes, barring bugs this is how it should work. I would really like that people would test more before they see the updated package in their favorite distro, but this never happens.

Issue #538, opened 2025-08-14
dkms add does not copy dot-files


ian-abbott:
My tarball contains a `.tarball-version` file in the top-level source directory, that is used to set the version number internally. However, dkms's `add_source_tree()` function uses the command `cp -fr "$from"/* "$source_tree/$module-$module_version"`, which does not match the dot-files in the `"$from"` directory.

I suggest changing it to `cp -fr "$from"/{,.[^.]}* "$source_tree/$module-$module_version"`. It could also be changed to `cp -frT "$from" "$source_tree/$module-$module_version"`, but POSIX `cp` does not support the `-T` option. (GNU coreutils `cp` and BusyBox `cp` both support it.)

ian-abbott:
It would also be nice if the above `cp` command preserved timestamps.

ian-abbott:
I realized that the `{,.[^.]}*` glob will miss double-dot-files, so `{,.[^.],..?}*` would be better.

PR #540, opened 2025-08-19
dkms: Source copy changes #538

ian-abbott:
The "Use cp's -T option instead of wildcards when copying source directory" commit implements #538 to copy dot-files. Since dkms is already using the `-a` option, I thought it would be fine to use the `-T` option supported by GNU coreutils cp and BusyBox cp.

The "Preserve timestamps, file modes, and ownership when copying source directory" additionally uses cp's `-p` option, mostly to preserve the timestamps. It may help in cases where the source directory contains generated sources that should not need rebuilding.

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -2429,7 +2429,7 @@ make_tarball()
         echo ""
         echo "Marking $source_dir for archiving..."
         mkdir -p "$temp_dir_name/dkms_source_tree"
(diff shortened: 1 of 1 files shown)


scaronni:
Of all the distributions that are part of the tests, none uses busybox, so I think it's safe to add.

PR #543, opened 2025-08-26
sparse arrays and a few more fixes

anbe42:
properly support (or rather error out on) sparse module arrays
use cp --reflink=auto
revert dkms.service installation on Debian


anbe42:
@scaronni:  I have one more change in the Debian package: adding back documentation for the deprecated CLEAN and the new special case CLEAN=true. Should we upstream that as well? Better wording welcome.

scaronni:
Hi @anbe42, this pull request looks great to me, let's merge it when you think you're ready, I would like to cut out a new release.

I've seen the patches for adding back the documentation for `CLEAN` and `CLEAN=true`. If it's ok for you, I would prefer to keep things as they are and not include them. The thing is deprecated, and that looks sort of like rolling it back.

I also made everyone switch to the new format here in NVIDIA for the dozen of DKMS packages, and would like to avoid having things lingering in various `dkms.conf` files forever.

Thanks.

PR #583, opened 2026-03-11
Allow patches, scripts and overrides in /etc/dkms

### LINKED ISSUE #562: [Feat. req.] Support patches and scripts in `/etc/dkms/`
Currently it's mandated that patches and scripts live in the root directory of the source, i.e. `/usr/src/mod-ver/`. No other locations are supported. However this is somewhat of a problem if `/usr/src/mod-ver/` is managed by a package manager. Package managers usually make no guarantees that files placed here won't be deleted or overwritten, so manually changing stuff here is generally frowned upon.

Considering that there's already support for `dkms.conf` override in `/etc/dkms/`, I think it'd be a good idea to extend this a bit so that custom patches and scripts can be placed here too.

scaronni:
Resolves https://github.com/dkms-project/dkms/issues/562.

Overrides in `/etc/dkms`:

- Patches declared with the `PATCH[#]` directives, can now:
  - Reside in `/etc/dkms/<module>/patches/`
  - Be overridden by a patch with the same name in `/etc/dkms/<module>/patches/`
  - As before,  the directory structure is preserved, as the whole `patches` directory is copied.
- Scripts declared with `POST_ADD`, `PRE_BUILD`, `POST_BUILD`, `PRE_INSTALL`, `POST_INSTALL` and `POST_REMOVE` can now:
  - Reside in `/etc/dkms/<module>/`
  - Be overridden by a script with the same name in `/etc/dkms/<module>/`

### Overriding a patch

Original:
```
$ ls -l /usr/src/gasket-1.0/my.patch
-rw-r--r--. 1 root root 2909 Mar  5 01:00 /usr/src/gasket-1.0/my.patch
$ cat /usr/src/gasket-1.0/dkms.conf
PATCH[0]="my.patch"
```
Override:

```
$ ls -l /etc/dkms/gasket/patches/my.patch
-rw-r--r--. 1 root root 2828 Mar 16 01:00 /etc/dkms/gasket/patches/my.patch
```

### New local patch

Adding a new local patch that survives across DKMS module upgrades:

```
$ ls -l /etc/dkms/cxr88/patches/local.patch
-rw-r--r--. 1 root root 2834 Mar 19 01:00 /etc/dkms/cxr88/patches/local.patch
$ cat /etc/dkms/cxr88.conf
PATCH[0]="local.patch"
```

### New post script

Adding a new post script that survives across DKMS module upgrades:

```
$ ls -l /etc/dkms/nvidia/scripts/reload-nvidia-peermem.sh
-rw-r--r--. 1 root root 2834 Mar 19 01:00 /etc/dkms/nvidia/scripts/reload-nvidia-peermem.sh
$ cat /etc/dkms/nvidia.conf
POST_INSTALL="reload-nvidia-peermem.sh"
```

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -58,6 +58,8 @@ TEST_MODULES=(
 )
 TEST_TMPDIRS=(
     "${tmpdir}/dkms_test_dir_${KERNEL_VER}/"
+    "/etc/dkms/dkms_patches_test/"
+    "/etc/dkms/dkms_scripts_test/"
 )
 TEST_TMPFILES=(
     "${tmpdir}/dkms_test_private_key"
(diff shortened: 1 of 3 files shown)


scaronni:
@cyqsimon can I get some feedback? To test:

```
$ git clone https://github.com/dkms-project/dkms.git
$ cd dkms
$ make
$ sudo cp dkms $(which dkms)
```

run_test.sh
```diff
@@ -2530,6 +2532,71 @@ run_status_with_expected_output 'dkms_scripts_test' << EOF
 dkms_scripts_test/1.0: added
 EOF
 
+echo 'Testing patches override from /etc/dkms'
+mkdir -p /etc/dkms/dkms_patches_test/patches
+cp test/dkms_patches_test-1.0/patches/patch1.patch /etc/dkms/dkms_patches_test/patches/patch1.patch
+run_with_expected_output dkms install -k "${KERNEL_VER}" -m dkms_patches_test -v 1.0 << EOF
+${SIGNING_PROLOGUE}
+Applying patch patch1.patch... done.
+Applying patch subdir/patch2.patch... done.
```
zeroepoch: I'm assuming `/etc/dkms/<module>/patches/subdir/patch2.patch` overrides the system `subdir/patch2.patch`? Meaning the `subdir` is also respected and checked? I wonder if the test should be reversed? Seems like more of a nitpick though.

zeroepoch (approved):

run_test.sh
scaronni: It's a good point, I'll amend the tests for this use case. Probably also the man page.

amilonenv (approved):
LGTM

PR #587, opened 2026-05-07
Add deferred logic for BUILD_DEPENDS

scaronni:
Add deferred logic for `BUILD_DEPENDS` with multiple package upgrades (kernel, modules and dependent modules) when invoking `autoinstall`. 

### The issue

Practical issue in Ubuntu:

1. Module with a dependency specified:
```
$ cat /usr/src/nvidia-fs-2.26.6/dkms.conf | grep BUILD_DEPENDS
BUILD_DEPENDS[0]="nvidia"
```

2.  A transaction involving an update of the kernel and the `nvidia` DKMS module at the same time:
```
Autoinstall on 6.14.0-1011-nvidia-64k succeeded for module(s) kernel-mft-dkms knem mlnx-ofed-kernel xpmem.
nvidia-fs/2.26.6 autoinstall failed due to missing dependencies: nvidia.

Error! One or more modules failed to install during autoinstall.
Refer to previous errors for more information.
run-parts: /etc/kernel/header_postinst.d/dkms exited with return code 1
dpkg: error processing package linux-headers-6.14.0-1011-nvidia-64k (--configure):
 installed linux-headers-6.14.0-1011-nvidia-64k package post-installation script subprocess returned error exit status 1
```

3. At the end of the upgrade, DKMS triggers a rebuild of `nvidia-fs`, which now succeeds because the `nvidia` module has now been built:
```
Autoinstall of module nvidia-fs/2.26.6 for kernel 6.14.0-1011-nvidia-64k (aarch64)
Building module(s).......... done.
Signing module /var/lib/dkms/nvidia-fs/2.26.6/build/nvidia-fs.ko
Found pre-existing /lib/modules/6.14.0-1011-nvidia-64k/ubuntu/nvidia-fs/nvidia-fs.ko.zst, archiving for uninstallation
Installing /lib/modules/6.14.0-1011-nvidia-64k/updates/dkms/nvidia-fs.ko.zst
Running depmod... done.

Autoinstall on 6.14.0-1011-nvidia-64k succeeded for module(s) iser isert kernel-mft-dkms knem mlnx-nfsrdma mlnx-nvme mlnx-ofed-kernel nvidia srp xpmem nvidia-fs.
```

But at this point, the initial failure causes the `apt` transaction to report a failure:
```
Errors were encountered while processing:
 linux-headers-6.14.0-1011-nvidia-64k
 linux-headers-nvidia-64k-6.14
 linux-nvidia-64k-6.14
```

### Deferred logic for hard dependencies

`autoinstall` now distinguishes two kinds of unresolved `BUILD_DEPENDS` (taken from the comments in `dkms.in`):

- "real" missing dependency:
  The dependency is a DKMS-managed module (in `known_modules`) that we tried but could not install (e.g. circular dependency, build failure of the dependency).

- "deferred" missing dependency:
  The dependency is not registered with DKMS at all. This commonly happens during multi-package installs where the dependency's source package has not yet been configured by `dpkg/rpm`, so its `postinst` hasn't run `dkms add` yet. The dependency may also be supplied by something outside DKMS entirely. Either way, the current `autoinstall` run  cannot make progress on it, but a later trigger (the dependency package's own `postinst`, or a re-run of this hook) will. Don't fail the whole transaction in that case.

By running `dkms build` directly, as it was before, exit 13 is still returned unless `--force` is used. The relaxation (the deferred logic) is only on the automated path where DKMS doesn't have full visibility into pending package operations.

### DIFF
--- a/run_test.sh
+++ b/run_test.sh
@@ -1833,12 +1833,12 @@ run_status_with_expected_output 'dkms_dependencies_test' << EOF
 dkms_dependencies_test/1.0: added
 EOF
 
-echo "Running dkms autoinstall (expected error)"
-run_with_expected_error 11 dkms autoinstall -k "${KERNEL_VER}" << EOF
-dkms_dependencies_test/1.0 autoinstall failed due to missing dependencies: dkms_test.
-
-Error! One or more modules failed to install during autoinstall.
-Refer to previous errors for more information.
+echo "Running dkms autoinstall (expected deferred, BUILD_DEPENDS not in DKMS)"
+run_with_expected_output dkms autoinstall -k "${KERNEL_VER}" << EOF
+dkms_dependencies_test/1.0 autoinstall deferred: BUILD_DEPENDS not registered with DKMS: dkms_test.
+  These may be provided externally or by a package not yet configured;
+  autoinstall will retry on the next trigger.
+Warning: Autoinstall on ${KERNEL_VER} deferred module(s) dkms_dependencies_test/1.0 due to BUILD_DEPENDS not yet registered with DKMS.
 EOF
 run_status_with_expected_output 'dkms_test' << EOF
 EOF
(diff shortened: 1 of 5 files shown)


amilonenv (approved):
It looks good to me.

run_test.sh
```diff
@@ -1833,12 +1833,11 @@ run_status_with_expected_output 'dkms_dependencies_test' << EOF
 dkms_dependencies_test/1.0: added
 EOF
 
-echo "Running dkms autoinstall (expected error)"
-run_with_expected_error 11 dkms autoinstall -k "${KERNEL_VER}" << EOF
-dkms_dependencies_test/1.0 autoinstall failed due to missing dependencies: dkms_test.
-
-Error! One or more modules failed to install during autoinstall.
-Refer to previous errors for more information.
+echo "Running dkms autoinstall (expected deferred, BUILD_DEPENDS not in DKMS)"
```
zeroepoch, line 1837: Should a negative test be added that is expected to fail due to a non-deferred dependency that is missing?

zeroepoch (approved):

run_test.sh
scaronni: We have already the following tests:
- `dkms_circular_dependencies_test` - self loop, exit 11
- `dkms_failing_dependencies_test` - dep is in DKMS, but build fails, exit 11
- `dkms_dependencies_test` - deferred, exit 0

scaronni: I've added this as a negative test, the tests are currently running:
- `dkms_dependencies_on_noautoinstall_test` - failure, exit 11, the dependency is in DKMS but has no `AUTOINSTALL` so it results in the dependency not being found.

scaronni: Seems to be fine, marking resolved.

PR #594, opened 2026-07-13
dkms: Use a generic missing kernel headers error message

HadiHbus:
The missing kernel headers error only suggested the Debian-style linux-headers package.
Add the rpm-based kernel-devel package to make the DKMS guidance useful across packaging ecosystems, and update the expected test output accordingly.


scaronni:
True, but I would remove the mention completely about the speciific debian package and just use a generic sentence:
- `kernel-devel-${KERNEL_VER}-noheaders` does not exist in most distributions (RHEL/Fedora), it's usually `kernel-devel`.
- SUSE has completely different package names
- All the other distributions are not covered (Arch, Gentoo, etc.).

PR #597, opened 2026-07-17
Adjust module location matches

### LINKED ISSUE #595: RHEL 10 not matched by `override_dest_module_location()`
Contents of `/etc/os-release`:
* `ID_LIKE="fedora"` in RHEL 7,8,9 
* `ID_LIKE="centos fedora"` in RHEL 10

DKMS v3.4.1/HEAD only knows `fedora*` and `rhel*` but not `centos*`. As a result, `DEST_MODULE_LOCATION` is overridden in RHEL < 10 but preserved in RHEL >= 10. This does not appear to be intended behaviour.

Isolated reproducer:
```console
echo '
distro_version()
{
    for f in /etc/os-release /usr/lib/os-release; do
        if [[ -e $f ]]; then
            (
                # shellcheck disable=SC1090
                . "$f"
                if [[ "$ID" = "ubuntu" ]]; then
                    # ID_LIKE=debian in ubuntu
                    echo "$ID"
                elif [[ ${#ID_LIKE[@]} != 0 ]]; then
                    echo "${ID_LIKE[0]}"
                else
                    echo "$ID"
                fi
            )
            return
        fi
    done
    die 4 "System is missing os-release file."
}
override_dest_module_location()
{
    local orig_location
    orig_location="$1"
    [[ ${addon_modules_dir} ]] && echo "/${addon_modules_dir}" && return

    case "$running_distribution" in
    fedora* | rhel* | ovm*)
        echo "/extra" && return
        ;;
    sles* | suse* | opensuse*)
        echo "/updates" && return
        ;;
    debian* | ubuntu*)
        echo "/updates/dkms" && return
        ;;
    arch*)
        echo "/updates/dkms" && return
        ;;
    *)
        ;;
    esac
    echo "$orig_location"
}
running_distribution=$(distro_version) && echo $running_distribution
override_dest_module_location /orig_location
' > dkms_distro_version.sh

$ cat dkms_distro_version.sh | docker run --rm -i redhat/ubi9 bash
fedora
/extra

$ cat dkms_distro_version.sh | docker run --rm -i redhat/ubi10 bash
centos fedora
/orig_location
```

Perhaps `*fedora*` would be suitable (which would [also](https://github.com/search?q=repo%3Achef%2Fos_release%20%2FID_LIKE.*fedora%2F&type=code) cover CentOS, AlmaLinux, Rocky, etc in place of `rhel*`) or `distro_version()` could favour known distros.

scaronni:
Fixes https://github.com/dkms-project/dkms/issues/595.

Be more explicit with distribution matching for the Red Hat case: `fedora/centos/rhel`.

Drop `ovm`, it's dead now.

### DIFF
--- a/dkms.in
+++ b/dkms.in
@@ -513,7 +513,7 @@ override_dest_module_location()
     [[ ${addon_modules_dir} ]] && echo "/${addon_modules_dir}" && return
 
     case "$running_distribution" in
-    fedora* | rhel* | ovm*)
+    fedora* | centos* | rhel*)
         echo "/extra" && return
         ;;
     sles* | suse* | opensuse*)


amilonenv (approved):
It looks good to me.

pcolledg-amd (approved):

